PR #1, opened 2017-06-30
Add hash router capability to router.js

jrop:
Abstract out core routing functionality and construct two routers for export at module load:

* The usual `Router`
* `Router.Hash` that uses hash routing.

The interface for both is equivalent and can be used like the following:

```js
const {app, Router} = hyperapp
app({
  mixins: [Router]
})
// or
app({
  mixins: [Router.Hash]
})
```

The rollup file `hyperapp.js` increases by only 238 bytes with this addition:

```txt
-rw-r--r--  1 me  me   3621 Jun 29 11:39 hyperapp.js (before)
-rw-r--r--  1 me  me   3859 Jun 29 11:39 hyperapp.js (after)
```

Prior work:
* hyperapp/hyperapp/pull/79
* ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "bundle": "rollup -i src/index.js -o dist/router.js -f umd -mn router",
     "minify": "uglifyjs dist/router.js -o dist/router.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --source-map dist/router.js.map",
     "prepublish": "npm run build",
-    "format": "prettier --semi false --write 'src/**/*.js'",
+    "format": "prettier --semi false --write '{src,test}/**/*.js'",
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
   },
   "babel": {
@@ -39,6 +39,7 @@
     "babel-preset-es2015": "^6.24.1",
     "hyperapp": "^0.12.0",
     "jest": "^20.0.4",
+    "jest-cli": "^20.0.4",
     "prettier": "~1.5.3",
     "rollup": "^0.47.6",
     "uglify-js": "^2.7.5"
--- a/src/router.js
+++ b/src/router.js
@@ -1,4 +1,17 @@
+function isHistoryApiAvailable() {
+  return window.history && window.history.pushState
+}
+
 export function router(options) {
+  options = options || {}
+  if (typeof options.hash == "undefined")
+    options.hash = !isHistoryApiAvailable()
+
+  function getLocation() {
+    return options.hash
+      ? (location.hash || "/").replace(/^#/, "")
+      : location.pathname
+  }
   return function(emit) {
     return {
       state: {
@@ -12,8 +25,9 @@ export function router(options) {
             }
           },
           go: function(state, actions, path) {
-            if (location.pathname + location.search !== path) {
-              history.pushState({}, "", path)
+            if (getLocation() !== path) { // TODO: what about location.search?
+              if (options.hash) location.hash = path
+              else history.pushState({}, "", path)
               actions.router.set({
                 path: path
               })
@@ -23,17 +37,19 @@ export function router(options) {
       },
       events: {
         load: function(state, actions) {
-          addEventListener("popstate", function() {
-            actions.router.set({})
-          })
+          addEventListener(
+            options.hash ? "hashchange" : "popstate",
+            function() {
+              actions.router.set({})
+            }
+          )
         },
         render: function(state, actions, view) {
           return view[
             (state.router.index >= 0
               ? state
-              : actions.router.set(
-                  emit("route", match(location.pathname, view))
-                )).router.index
+              : actions.router.set(emit("route", match(getLocation(), view))))
+              .router.index
           ][1]
         }
       }
--- a/src/router.js
+++ b/src/router.js
@@ -25,7 +25,8 @@ export function router(options) {
             }
           },
           go: function(state, actions, path) {
-            if (getLocation() !== path) { // TODO: what about location.search?
+            if (getLocation() !== path) {
+              // TODO: what about location.search?
               if (options.hash) location.hash = path
               else history.pushState({}, "", path)
               actions.router.set({
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -10,6 +10,7 @@ Object.defineProperty(window.location, "pathname", {
 beforeEach(() => {
   document.body.innerHTML = ""
   location.pathname = "/"
+  location.hash = ""
   history.pushState = function(state, title, url) {
     location.pathname = url
   }
@@ -404,3 +405,134 @@ test("fire route only if path changes", done => {
   })
 })
 
+test("{ hash: true }", done => {
+  app({
+    view: [
+      [
+        "/",
+        () =>
+          h(
+            "div",
+            {
+              oncreate() {
+                expect(document.body.innerHTML).toBe(`<div>foo</div>`)
+                done()
+              }
+            },
+            "foo"
+          )
+      ]
+    ],
+    mixins: [router({ hash: true })]
+  })
+})
+
+test("#/ ({ hash: true })", done => {
+  location.hash = "#/"
+  app({
+    view: [
+      [
+        "/",
+        () =>
+          h(
+            "div",
+            {
+              oncreate() {
+                expect(document.body.innerHTML).toBe(`<div>foo</div>`)
+                done()
+              }
+            },
+            "foo"
+          )
+      ]
+    ],
+    mixins: [router({ hash: true })]
+  })
+})
+
+test("#/about ({ hash: true })", done => {
+  location.hash = "#/about"
+  app({
+    view: [
+      ["/", () => h("div", "foo")],
+      [
+        "/about",
+        () =>
+          h(
+            "div",
+            {
+              oncreate() {
+                expect(document.body.innerHTML).toBe(`<div>bar</div>`)
+                done()
+              }
+            },
+            "bar"
+          )
+      ]
+    ],
+    mixins: [router({ hash: true })]
+  })
+})
+
+test("go ({ hash: true })", done => {
+  app({
+    view: [
+      [
+        "/",
+        (state, actions) =>
+          h("div", {
+            oncreate() {
+              expect(document.body.innerHTML).toBe("<div></div>")
+              actions.router.go("/foo")
+            }
+          })
+      ],
+      [
+        "/foo",
+        (state, actions) =>
+          h(
+            "div",
+            {
+              onupdate() {
+                expect(location.hash).toBe("#/foo")
+                expect(document.body.innerHTML).toBe("<div>foo</div>")
+                actions.router.go("/bar")
+              }
+            },
+            "foo"
+          )
+      ],
+      [
+        "/bar",
+        (state, actions) =>
+          h(
+            "div",
+            {
+              onupdate() {
+                expect(location.hash).toBe("#/bar")
+                expect(document.body.innerHTML).toBe("<div>bar</div>")
+                actions.router.go("/baz")
+              }
+            },
+            "bar"
+          )
+      ],
+      [
+        "/baz",
+        (state, actions) =>
+          h(
+            "div",
+            {
+              onupdate() {
+                expect(location.hash).toBe("#/baz")
+                expect(document.body.innerHTML).toBe("<div>baz</div>")
+                done()
+              }
+            },
+            "baz"
+          )
+      ]
+    ],
+    mixins: [router({ hash: true })]
+  })
+})


jbucaran:
@jrop Why two routers?

jrop:
@jbucaran Yes, mostly because of size.  There is reusable logic in Router that does path-matching and I wanted to leverage that.  To be fair there are numerous ways this could be accomplished:

1. The way this PR does it
2. One router that takes an option, e.g. `Router({hash: true})` (defaults to false)
3. Autodegradation in browsers that do not support the history API: if the history API is not available, use hashes

Even with my approach, I have exported it attached to the `Router` object (to be used like `Router.Hash`.  However, it could also be its own export:

```js
export {Router, HashRouter}
```
I don't like option (3) because in my case, I need to use hash routing sometimes no matter what the browser supports, and (1 or 2) would give me the control I need to accomplish this.  Even with option (1 or 2), I can write my own gracefully degrading router logic quite easily:

```js
const {app, Router} = hyperapp
app({
  mixins: [browserSupportsHistoryAPI() ? Router : Router.Hash]
})
```

Option 2 may be more user-friendly though; I could adjust my PR to accomodate an API more like:

```js
const {app, Router} = hyperapp
app({
  mixins: [Router({hash: true})]
})
```

I submitted as is mostly to spark discussion (which is also why I deferred adding tests at this point).

jbucaran:
I vote for option (3) in "official" router, but (2) is not bad at all. What do you think? @lukejacksonn

jrop:
How about a mix of (2) and (3):

```js
{
  mixins: [Router()] // note the function call instead of just [Router]: will prefer history API and fallback to hash

  mixins: [Router({hash: false})] // will force history API

  mixins: [Router({hash: true})] // will force hash router
}
```

I would prefer this: it guesses at sane defaults, but allows for fine-grained control.

jbucaran:
> @jrop I don't like option (3) because in my case, I need to use hash routing sometimes no matter what the browser supports.

Can I ask why?

jrop:
@jbucaran Yeah.  In my case sometimes I need to create one-off pages that are not full-blown apps.  I have found it to be convenient to use a small framework like hyperapp, and I can deploy these mini-apps as a single HTML page by uploading them to our WordPress site.  The page can then be accessed at a URL like `http://domain.tld/wp-content/uploads/.../my-page.html`.  Now, suppose I pushState '/about', my URL will become `http://domain.tls/about`.  Hyperapp can handle this, but if a user bookmarks that location, it will not be served up by Apache, which is problematic.  Hence I want to use hash routing no matter what.

I've been thinking about this and playing with it, and I think the following is important:

1. Maintian backwards compatibility: allow `mixins: [Router]`
    * Prefers the history API, falls back to the hash router
2. Allow configurability with `mixins: [Router.configure({hash: false})]`
    * This is important to me since I need a way to turn off the history API router, even if the history API is available

I have this coded up and will commit it awaiting your approval.

jbucaran:
@jrop Oh, but you can always use [hyperapp-hash-router](https://www.npmjs.com/package/hyperapp-hash-router). 😊😬

jrop:
@jbucaran True that (if by that you mean [zaceno/hyperapp-hash-router](https://github.com/zaceno/hyperapp-hash-router)).

An argument could be made that adding it to this module requires minimal extra code, but I see your point.

@zaceno Want to weigh in on this?

jbucaran:
@jrop Nice find! I didn't know. I don't want to make light of your use case, but I don't believe the official router should go out of his way to support this use case either when we can simply have a HashRouter that an interested party in the community (for example yourself or @zaceno, etc.) can maintain.

jrop:
@jbucaran Fair enough.  Then perhaps I will look into collaborating with @zaceno.  For the record I just pushed up the change that I was talking about.  Thanks for hyperapp!  It's truely an impressive little framework 😄

jbucaran:
@jrop In the meantime, perhaps you'd be interested (if I can convince you) in adding hash-based routing resolution to this router. 😉😉

jrop:
@jbucaran How exactly do you mean?  Do you mean supporting hash fragments in this router (not to be confused with storing the route in the hash)?

jbucaran:
@jrop Like described in the README (I just added this very recently).

> @hyperapp/router provides actions and events for routing client-side pages with HyperApp. For browsers which don't yet support the History API, the router handles transparent translation to the fragment version of the URL.

jrop:
@jbucaran Coincidentally, this PR supports gracefully degrading to hash routing as your README mentions (take a look at the latest commit I pushed up).  At the expense of perhaps 1-5loc it also adds the `Router.configure(options)` method that supports my use case as well 😉 .

jbucaran:
Thanks @jrop! I'll think about this again later! 🙇😉😊

zaceno:
Hi sorry I'm late to the party! I'm happy to give up the name "hyperapp-hash-router", and will happily help - but I don't feel strongly about being involved either.

I just whipped up that repo as a proof of concept to share. I guess I thought I might make a proper repo of it at some
point, but I forgot about it after I started adopting a pattern of mapping routes to state rather than to views.

lukejacksonn:
@jrop thanks for all the work and well thought out points, you have highlighted some really important considerations. Sorry I've been a while getting back to this!

A quick summary:

- Now that the router is not built-into hyperapp, we have more freedom to do and undo.
- There is reusable logic in Router that does path-matching 
- One router that takes an option
- Graceful fallback to # in browsers that do not support the history API
- I can write my own gracefully degrading router logic quite easily
- I need to use hash routing sometimes no matter what the browser supports
- Maintain backwards compatibility
- Allow configurability with mixins

Those things considered I think:

1. Mixins in general should only export one const

Otherwise as @jbucaran says, why not just create `hyperapp-hash-router` and export `HashRouter` as a mixin. I appreciate the concern for there becoming _hyperapp-another-duplicate-mixin-2_ but this is not specific to the router and impossible to enforce. Anyhow, we should be encouraging people to create mixins (if mixins are here to stay).

2. Mixins being functions and taking options is good

Like `Router({ hash: true })`. Passing some config to a mixin is going to be pretty common. A lot of middleware works like this anyway. Maybe we encourage (or even insist) that mixins are functions that export `{state, actions,...}`. I'm not convinced on `mixin. configure` 😕 despite making it backward compatible.

3. Hash based routing is not just for old browsers

I had never really considered supporting has based routing out of the box.. but thinking about it now. I can start to appreciate how it might make hyperapp in general more adoptable. Take for instance cases where users don't have full control over the location pathname. Like on wordpress as mentioned, or github pages, or codepen (would this make routing possible on codepens :thinking:).

That is all I have got so far.. I am in no rush to merge this is just yet as this could be done by using @zaceno's or your own `hyperapp-hash-plugin` plugin. Regardless, this has been a super useful and I am certainly swaying toward supporting this is the _official_ router as supposed to `yet-another-router-mixin`.

> I started adopting a pattern of mapping routes to state rather than to views.

@zaceno that sounds like some forward thinking ⚡️😉

jbucaran:
@lukejacksonn Since we don't support IE9, there is no scenario in which gracefully degrading the router makes sense. Would it be possible to check if the user is using a hash in the route key and if so, use hash based routing?

jrop:
@lukejacksonn 
> Mixins being functions and taking options is good

I agree. The reason I added `.configure(...)` to the PR was to maintain backward compatibility.  I agree that this is slightly kludgey, and I would also prefer to remove it in order to keep this as clean as possible.

> Hash-based routing is not just for old browsers...would this make routing possible on codepens 🤔?

The pen is embedded in its own iframe, so conceivably hash-based routing would work with codepen.

> Take for instance cases where users don't have full control over the location pathname

I think this happens more often than not, which is why I agree that hash-based routing should ship out-of-the-box.  In many cases, I do not have the luxury of controlling the entire target deployment environment (most development ends up hacking an existing system).  In this respect, hash-based routing is more adaptable.

@jbucaran 
> Since we don't support IE9, there is no scenario in which gracefully degrading the router makes sense

Good point.  I think this then brings the discussion back to the point-of-view that hash-routing is not a degraded mode of operation, but one that supports more environments (like when the paths cannot be controlled).

> Would it be possible to check if the user is using a hash in the route key and if so, use hash based routing?

You mean like this?

```js
app({
  views: [
    ['#/', () => <Home />],
    ['#/about', () => <About />],
  ]
})
```

I find this intriguing, but on another note, I think implementing this would require a lot more code than the current implementation (that is just a guess, though).  Would it also allow potentially strange mixing of hash vs. non-hash routing?

jbucaran:
> @jrop You mean like this?

Yes, like that. I like that because to me, hash-based routing is like an eccentric ice flavor. If you really want it, then accept what you are given kind of approach. In a good and well-manered way, of course! 😄 

>  I think implementing this would require a lot more code than the current implementation.

Well, if I can get some PRs done tonight in HyperApp base, I'll give this one a shot. If you want to go first, then be my guest.

jrop:
@jbucaran I guess I just have more questions about how that would behave.  I can think of an ambiguous situation:

```js
app({
  views: [
    ['/about', () => <About />],
    ['/#contact', () => <Contact />],
  ]
})
```

How would the route matcher resolve a path such as `/about#/contact`?  Technically this path matches both routes.

jbucaran:
@jrop Wouldn't it work if we look at any of the routes during initialization and if there is a `#/` then we turn hash mode on, otherwise use the history API?

Also, did you mean `/about#/contact` or `/about` and `#/contact`?

jrop:
@jbucaran I suppose you look at the routes and check for a `#`.  My point may be irrelevant now that I understand what you are saying, but was that `/about#/contact` is a valid route: it contains a normal path fragment _and_ a hash fragment:

```js
window.location.pathname = '/about'
window.location.hash     = '#/contact'
```

I was just trying to think of ambiguous situations where `location.pathname` would match a route, and `location.hash` might match another completely different route and how to handle that ambiguity.  I see what you are saying now, though.  You are saying, for example, check the first route, and if it contains `#/`, then assume that the user wants to run the router in hash-mode, otherwise use history.

I am opposed to this solution because, say I start coding my app using hash-routing, and then later in the applications development, I want to switch to history, I will have to change all of my routes versus just changing a configuration option.  I believe routing should be an abstract concept (path matching `/my/path/:param`) separated from its implementation (history/hash-fragments/state).

dodekeract:
Why do we need hash-based routing at all? `history` is superior and supported in all modern, relevant browsers.

zaceno:
@dodekeract I don't know for sure that hash based routes need to be supported in the *official* router, but I *do* know there is a *need* for hash based routes. In instances when you have no control over the file server serving the app (so you can't redirect 404's), when you have multiple html files (so you can't just naively redirect 404's to index.html), or if you want your app to be completely standalone (i e possible to use over `file:///` protocol)

jrop:
FYI, in the meantime for my personal use I have written [this module](https://github.com/jrop/hyperapp-routes) (hyperapp-routes).

jbucaran:
@jrop Let's do this! Now that the router is fixed with hyperapp@latest, we can go on with this.

jrop:
Alright, I have reapplied my changes on top of master.  Please let me know what you think.  You can test with this:

```js
<!DOCTYPE html>
<html>
  <body>
    <script src="hyperapp.js"></script>
    <script src="router.js"></script>
    <script>
      const {app, h} = hyperapp
      app({
        mixins: [Router.Router({hash: true})],
        view: [
          ['/', (_, actions) => h('div', null,
            h('div', null, 'Index'),
            h('a', {
              href: 'javascript: void(0)',
              onclick: () => actions.router.go('/about'),
            }, 'goto about'))],
          ['/about', (_, actions) => h('div', null,
            h('div', null, 'About'),
            h('a', {
              href: 'javascript: void(0)',
              onclick: () => actions.router.go('/'),
            }, 'goto index'))]
        ]
      })
    </script>
  </body>
</html>

```

jbucaran:
@jrop Thanks, how does this change the current router usage?

Also, we'll need tests and docs if we want to merge this at some point. 😉

jrop:
@jbucaran Indeed!  I will need to add tests if this is in the approved final format.

In its current state, this changes usage from:

`mixins: [Router]`
to:
`mixins: [Router(options)]`

So I just want to confirm before I add tests: can I proceed with this design?

jbucaran:
@jrop So what would be the options for when not using hash?

jbucaran:
@jrop What about my previous idea about defaulting to hash-based routing if you used "#foo" or "#bar" instead of "/foo" or "/bar"?

/cc @lukejacksonn @dodekeract

jrop:
@jbucaran To answer the first question, in order to use hash-based routing, one would do:
`mixins: [Router({hash: true})]`
To use the history API:
`mixins: [Router()]`, or `mixins: [Router({hash: false})]`

As for the second question, you are proposing auto-detecting hash-based routing if all of the defined routes start with a hash, is that correct?  So if I define routes like:

```js
app({
  view: [
    ['/', () => ...],
    ['/foo', () => ...],
  ],
  mixins: [Router]
})
```
...then I should assume that, by usage, history-based routing is preferred.

However, if the views are defined as:

```js
app({
  view: [
    ['#/', () => ...],
    ['#/foo', () => ...],
  ],
  mixins: [Router]
})
```
... then the library could assume hash-based routing?

Whatever is the consensus, I can start on that and then add tests.

jbucaran:
Sure, let's first all agree: /cc @SkaterDad @lukejacksonn @zaceno

zaceno:
I probably haven't thought enough about this to have an opinion, but I like the "dynamically switch to hash based if routes have hashes"-approach. It makes sense to me.

jbucaran:
Thanks @zaceno! I'd still like to hear how @SkaterDad and @lukejacksonn feel about this.

SkaterDad:
Automatic detection of hash routing is an interesting idea, so I'm supportive of that given the current router implementation.

I'm not sure how well that would work with the more dynamic component-style routing, though?  Could you mix & match hash & non-hash route matching?

Another thing that other routers do is fallback to hash routing when the environment doesn't support the history API.  Is that something we're not going to worry about since `hyperapp` doesn't support old browsers?

jbucaran:
> Is that something we're not going to worry about since hyperapp doesn't support old browsers?

Correct.

dodekeract:
@jbucaran Is there any reason why hash and normal routing would be mutually exclusive? If you have some routes with a `#`, I'd expect those routes to behave like hash routes. If you have some routes with a `/`, I'd expect those routes to behave like normal routes.

jbucaran:
@dodekeract You are probably right, but the approach we are going to take is basically what @jrop originally proposed (not the one I did).

```jsx
app({
  mixins: [router({ hash: true })]
})
```

So, you can't have both.

@jrop Sorry for the awful amount of waiting, but I am ready to move forward with this PR.

The good news is your early proposal was brought back to life!

lukejacksonn:
@dodekeract makes a good point but I'm happy with passing in a `hash: true` option also.. what  change are needed to get it merged @jrop ?

dodekeract:
@jbucaran @lukejacksonn I'm also fine with `{hash: true}`. Was just wondering if we need that at all.

jbucaran:
@dodekeract Well, @jrop needs it no?

dodekeract:
@jbucaran With "that" I mean the mixin option `{hash: true}`.

jrop:
@jbucaran @lukejacksonn  
Sounds good, then I think this is pretty close.  There are just a few things that I need to get done in order to have this ready:

- [ ] rebase onto master - switch to lower-case `router`
- [ ] add tests

I will try to work on that today if I have time and get some commits pushed up

jbucaran:
@dodekeract The default is `hash: false`, so I don't know what you mean.

dodekeract:
@jbucaran

Me **wondering if we need that at all**:

> Is there any reason why hash and normal routing would be mutually exclusive? If you have some routes with a #, I'd expect those routes to behave like hash routes. If you have some routes with a /, I'd expect those routes to behave like normal routes. — @dodekeract

> You are probably right, but the approach we are going to take is basically what @jrop originally proposed — @jbucaran

> I'm also fine with {hash: true}. Was just **wondering if we need that at all**. — @dodekeract

I hope this makes it less confusing.

jbucaran:
Thanks @jrop! Looking forward to that.

@dodekeract Still not sure. If you have an alternative API proposal, better to just write it.

dodekeract:
@jbucaran I don't have an API proposal except the one with `#` and `/` where we support both. As I said, I'm 100% fine with the solution we chose. My last 3 comments are not referring to any new idea, or proposal. I'm simply stating that I agree with the decision and don't think we need to change anything.

jbucaran:
Okay, sorry for the misunderstanding. 👍👍

JorgeBucaran:
@jrop I just published 0.4.1; it works with hyperapp@0.18.0. I think we should close here, but start a new discussion about how to bring a hash router to this new arch. Thanks for the work you put on this and all the waiting. 👋

willin:
mark

PR #6, opened 2017-08-01
changes needed to update router from 0.9.x api to 0.10.x

andyrj:
this all seems to work on my projects, just need to update docs to move config from app.view into the args like:

```js
Router([['/', <div>hello</div>]])
```


jbucaran:
Thanks @andyrj. Fixed things with #12. 🙇

PR #7, opened 2017-08-05
Update router.js

infinnie:
`"*"` is a constant and means the regular expression `/.*/`, where the two asterisks mean two completely different things.


jbucaran:
@infinnie But isn't the result going to be the same?

```js
if (route === "*") console.log("." + route) // => .*
if (route === "*") console.log(".*") // => .*
```

Or did I miss something?

jbucaran:
".*" landed in #12. :+1:

PR #9, opened 2017-08-09
Refactors router and fixes tests

lukejacksonn:
The router is currently out of sync with the latest release of hyperapp and as it stands as intended or as documented (or at all for that matter) 😅

Although this PR may have a substantial diff.. the router it is now functional again and with almost exactly the behaviours as before. Here are a few things that have changed:

- The router now only accepts `emit` as an argument, like a regular mixin
- The router's state now looks like:
```js
router: {
  url: "/users/luke/27",
  route: { 
    url: "/users/luke/27",
    match: "/users/:id/:age",
    view: (s,a,d) => h('h1', {}, 'Hello!')
    params: {
      id: "luke",
      age: "27"
    }
  }
}
```
- The attribute `router.url` is kept in sync with `location.pathname`
- The attribute `router.route` is updated from within render when router.url !== router.route.url
- the router now uses the `load` event rather than the `loaded` event
- All tests have been updated to account for the new state structure and `app` being async

I _think_ (or hope at least) that this implementation is no more/less performant than the old one but I have not tested it for that really.

There is a breaking change for anyone accessing the routers state directly.. for instance `state.router.match` will now be undefined but can still be accessed through `state.router.route.match`.

The app still emits a `route` event whenever a new route is entered.

That is about all. This is probably going to be my last attempt at this style of router before experimenting further with component based routers as discussed here https://github.com/hyperapp/hyperapp/issues/295

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -1,70 +1,68 @@
-export function Router(app, view) {
+export function Router(emit) {
   return {
-    state: {
-      router: match(location.pathname)
-    },
     actions: {
       router: {
-        match: function(state, actions, data, emit) {
+        update: function(state, actions, data) {
           return {
-            router: emit("route", match(data))
+            router: {
+              url: location.pathname,
+              route: data || {}
+            }
           }
         },
         go: function(state, actions, data) {
           history.pushState({}, "", data)
-          actions.router.match(data.split("?")[0])
+          actions.router.update()
         }
       }
     },
     events: {
-      loaded: function(state, actions) {
-        match()
-        addEventListener("popstate", match)
-
-        function match() {
-          actions.router.match(location.pathname)
-        }
+      load: function(_, actions) {
+        actions.router.update()
+        addEventListener("popstate", function() {
+          actions.router.update()
+        })
       },
-      render: function() {
-        return view
-      }
-    }
-  }
-
-  function match(data) {
-    for (var match, params = {}, i = 0, len = app.view.length; i < len; i++) {
-      var route = app.view[i][0]
-      var keys = []
-
-      if (!match) {
-        data.replace(
-          RegExp(
-            route === "*"
-              ? "." + route
-              : "^" +
-                  route
-                    .replace(/\//g, "\\/")
-                    .replace(/:([\w]+)/g, function(_, key) {
-                      keys.push(key)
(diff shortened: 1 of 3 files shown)


src/router.js
```diff
@@ -1,70 +1,68 @@
...
+          })[0]
+          var keys = route[0].split("/")
+          while (keys[count] && (values = regex.exec(pathname))) {
+            keys[count][0] === ":" && (params[keys[count].slice(1)] = values[1])
+            count++
           }
-        )
+          emit(
+            "route",
+            actions.router.update({
```
lukejacksonn: This might be better written as..
```
actions.router.update(emit('route', { ... }))
```

lukejacksonn:
Just realised I have not updated the docs.. will do if the PR makes it any further than this

jcubic:
Travis build failed.

lukejacksonn:
@jcubic ahh yes.. this will be because it is looking at `git://github.com/hyperapp/hyperapp.git` for the most recent version (0.11.0) of hyperapp. Which as I just discovered, in discussion with @jbucaran on slack, means that the module does not get built and thus there is no `dist` directory.. hence the Travis error `Couldn't find module hyperapp`

SkaterDad:
While you're at it, should `bundlesize` be added to the `test` script like in core?

lukejacksonn:
@SkaterDad perhaps.. although I'm wary of employing such stringent analysis of size when it comes to mixins. It might not be every mixins primary concern.. I'm not suggesting size be completely disregarded but sometimes too much focus on size can obfuscate the intention of the code and sometimes hinder progression.

That said, adding `bundlesize` is pretty trivial and unobtrusive.. maybe it should be on _all_ the things 😅 in the name of data!

SkaterDad:
> although I'm wary of employing such stringent analysis of size when it comes to mixins. It might not be every mixins primary concern.. I'm not suggesting size be completely disregarded but sometimes too much focus on size can obfuscate the intention of the code and sometimes hinder progression.

Completely agree, but it's nice information to have.  Surprisingly few packages report their bundle size in the README or elsewhere, but I pretty much always want to know.

jbucaran:
Thanks @lukejacksonn & @SkaterDad. Fixed it up with #12. 🙇👍

PR #12, opened 2017-08-17
Update router to hyperapp's latest.

jbucaran:
- No breaking changes.
- Add new .index property to route info object.


src/router.js
```diff
@@ -1,70 +1,84 @@
...
-      match: match,
-      params: params
-    }
+    var route = routes[i][0]
+    var keys = []
+
+    data.replace(
+      RegExp(
+        route === "*"
+          ? "." + route
```
lukejacksonn: Someone did say, how come this isn't hard coded to `? ".*"`

jbucaran: Saves 2 bytes, but it's okay, let's ".*".

src/router.js
```diff
@@ -1,70 +1,84 @@
...
-        match: function(state, actions, data, emit) {
+        update: function(state, actions, router) {
           return {
-            router: emit("route", match(data))
+            router: router
           }
         },
-        go: function(state, actions, data) {
-          history.pushState({}, "", data)
-          actions.router.match(data.split("?")[0])
```
lukejacksonn: Is this case covered in the refactor?

jbucaran: @lukejacksonn Nope. We need to decide what the behavior is going to be.

lukejacksonn: So urls with query strings should certainly not break the router.. I guess we have two options:

1. Split the url at `?` take the first item in the split array and extract matched parts as `router.params` object. Ignore the second item in split array.
2. Split the url at `?` take the first item in the split array and extract matched parts as `router.params` object **and** take the second item in split array and extract parts as `router.query` object.

PR #13, opened 2017-08-19
Add support for special characters (space, parens, %) within query params

### LINKED ISSUE #10: Route not matched when parameter includes "special" characters
considering these routes:
```
[
     ['/home', home],
     ['/query/:query', query],
     ['*', (state, actions) => {
         return (
             <b>{location.pathname}</b>
         );
     }]
]
```

when I navigate to `/query/Hot` hyperapp-router correctly matches the `query` route.
however when I navigate to `/query/Hot%20Pocket` it does not match the `query` route and instead falls back to the `*` route.

the same happens if there is `:`, `(`, or `)` in the query parameter whether or not the "special" characters are URI encoded.

chiefsmurph:
I completely understand if there's a preference to not `decodeURI` the route params.

https://github.com/hyperapp/router/compare/master...chiefsmurph:master#diff-15ee858ae5d5aa9b3e2846cfd1c4c466R66

This fixes #10

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -47,7 +47,6 @@ function match(pathname, routes) {
   for (var i = 0; i < routes.length && !match; i++) {
     var route = routes[i][0]
     var keys = []
-
     pathname.replace(
       RegExp(
         route === "*"
@@ -55,14 +54,18 @@ function match(pathname, routes) {
           : "^" +
             route.replace(/\//g, "\\/").replace(/:([\w]+)/g, function(_, key) {
               keys.push(key)
-              return "([-\\.%\\w]+)"
+              return "([-\\.%\\w\\(\\)]+)"
             }) +
             "/?$",
         "g"
       ),
       function() {
         for (var j = 1; j < arguments.length - 2; ) {
-          params[keys.shift()] = arguments[j++]
+          var value = arguments[j++]
+          try {
+            value = decodeURI(value)
+          } catch (_) {}
+          params[keys.shift()] = value
         }
         match = route
         index = i
--- a/test/router.test.js
+++ b/test/router.test.js
@@ -199,6 +199,52 @@ test("routes with dashes into a single param key", done => {
   })
 })
 
+test("route params with parentheses", done => {
+  window.location.pathname = "/Parenthesis(rhetoric)"
+  app({
+    view: [
+      [
+        "/:foo",
+        state =>
+          h(
+            "div",
+            {
+              oncreate() {
+                expect(document.body.innerHTML).toBe(`<div>Parenthesis(rhetoric)</div>`)
+                done()
+              }
+            },
+            state.router.params.foo
+          )
+      ]
+    ],
+    mixins: [Router]
+  })
+})
+
+test("route params with uri encoded string", done => {
+  window.location.pathname = "/Batman%20(1989%20film)"
+  app({
+    view: [
+      [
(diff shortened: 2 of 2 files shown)


chiefsmurph:
this fixes #10

andyrj:
Nice table of safe and unsafe uri characters here, https://perishablepress.com/stop-using-unsafe-characters-in-urls/

lukejacksonn:
I think this is related https://github.com/hyperapp/hyperapp/pull/230#issuecomment-307093576 to special chars in the url. 

Decoding the url is a separate issue but it kind of makes sense seeing as you are always guaranteed part of a url as an input.

chiefsmurph:
In case you are wondering what the use case was for this: I was working on an app based on Wikipedia and I wanted the user to be able to pass in a Wikipedia page name as a parameter.  Many pages have spaces and parenthesis in the page names.

lukejacksonn:
I was wondering! For clarification can you give us some actual examples of urls that don't work with the current router?

chiefsmurph:
`/query/Philosophical methodology`
`/query/Jennifer%20Lawrence`
`/query/Jerry Seinfeld (comedian)`
were the types of query params I was trying to pass from Wikipedia

src/router.js
```diff
@@ -47,22 +47,25 @@ function match(pathname, routes) {
...
         "g"
       ),
       function() {
         for (var j = 1; j < arguments.length - 2; ) {
-          params[keys.shift()] = arguments[j++]
+          var value = arguments[j++]
+          try {
+            value = decodeURI(value)
+          } catch (e) {}
+          params[keys.shift()] = value
```
jbucaran: @chiefsmurph Can we rewrite this as?

```js
try {
  var key = decodeURI(arguments[j++])
} catch (e) {}
  params[keys.shift()] = key
}
```

src/router.js
```diff
@@ -47,22 +47,25 @@ function match(pathname, routes) {
...
-
     pathname.replace(
       RegExp(
         route === "*"
           ? ".*"
           : "^" +
             route.replace(/\//g, "\\/").replace(/:([\w]+)/g, function(_, key) {
               keys.push(key)
-              return "([-\\.%\\w]+)"
+              return "([-\\.%\\w\\(\\)]+)"
```
jbucaran: @chiefsmurph What does `\\(\\)` do?

src/router.js
jbucaran: Hmmm, nevermind, your code is the best. Just use `_` for the catch please.

src/router.js
chiefsmurph: it is the necessary escaping for the open and close parens.
`\\` gets converted to a single `\` and then when the regexp is run that single `\` escapes the `(` and `)`.  (I followed the syntax for the `.`)

jbucaran: I see, so we want to escape both ( and ).

chiefsmurph:
When a consensus is made, do you have a preference on whether I should squash these commits?

jbucaran:
Sure, please squash the commits or I'll just do it when I merge it. 👍

jbucaran:
Thanks @chiefsmurph! 🎉

chiefsmurph:
Awesome!  My first contribution to Hyperapp!

jbucaran:
Welcome aboard! 🚀

PR #15, opened 2017-08-21
Link Component Upgrades

SkaterDad:
**Description**
Added some common checks to the <Link> component which I found in other routers.

To ensure default browser behaviors work, we should not navigate if:
 - meta, ctrl, alt, shift keys are pressed
 - non-primary mouse button/enter key
 - link target = "_blank"

**Tests**
Maintained 100% coverage.  Used jest's mock functions.

**Bundle size**
Added ~50 bytes gzipped to the build.

**Simple example:**
https://www.webpackbin.com/bins/-Ks531DrGloAvqjQO26i

**Other potential enhancements**
Some other libraries, like react-router, allow user-defined `onclick` handlers on their link component.  They save a reference to the function and call it in the "final" click handler, then check for `e.defaultPrevented`.   Do you want this feature also?

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "router"
   ],
   "scripts": {
-    "test": "jest --coverage --no-cache",
+    "test": "jest --coverage --no-cache --testURL https://hyperapp.js.org",
     "build": "npm run bundle && npm run minify",
     "bundle": "rollup -i src/index.js -o dist/router.js -f umd -mn Router",
     "minify": "uglifyjs dist/router.js -o dist/router.js --mangle --compress warnings=false --pure-funcs=Object.defineProperty -p relative --source-map dist/router.js.map",
--- a/src/Link.js
+++ b/src/Link.js
@@ -5,12 +5,18 @@ export function Link(props, children) {
   props.to = null
 
   props.onclick = function(e) {
-    if (e.metaKey || e.altKey || e.ctrlKey || e.shiftKey) return
+    if (
+      e.button !== 0 ||
+      e.metaKey ||
+      e.altKey ||
+      e.ctrlKey ||
+      e.shiftKey ||
+      props.target === "_blank"
+    )
+      return
 
-    if (e.button === 0) {
-      e.preventDefault()
-      props.go(props.href)
-    }
+    e.preventDefault()
+    props.go(props.href)
   }
 
   return h("a", props, children)
--- a/src/Link.js
+++ b/src/Link.js
@@ -4,9 +4,13 @@ export function Link(props, children) {
   props.href = props.to
   props.to = null
 
-  props.onclick = function(element) {
-    element.preventDefault()
-    props.go(props.href)
+  props.onclick = function(e) {
+    if (e.metaKey || e.altKey || e.ctrlKey || e.shiftKey) return
+
+    if (e.button === 0) {
+      e.preventDefault()
+      props.go(props.href)
+    }
   }
 
   return h("a", props, children)
--- a/src/Link.js
+++ b/src/Link.js
@@ -12,8 +12,9 @@ export function Link(props, children) {
       e.ctrlKey ||
       e.shiftKey ||
       props.target === "_blank"
-    )
+    ) {
       return
+    }
 
     e.preventDefault()
     props.go(props.href)
--- a/src/Link.js
+++ b/src/Link.js
@@ -4,19 +4,19 @@ export function Link(props, children) {
   props.href = props.to
   props.to = null
 
-  props.onclick = function(e) {
+  props.onclick = function(event) {
     if (
-      e.button !== 0 ||
-      e.metaKey ||
-      e.altKey ||
-      e.ctrlKey ||
-      e.shiftKey ||
+      event.button !== 0 ||
+      event.metaKey ||
+      event.altKey ||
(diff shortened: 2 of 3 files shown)


src/Link.js
```diff
@@ -4,8 +4,18 @@ export function Link(props, children) {
   props.href = props.to
   props.to = null
 
-  props.onclick = function(element) {
-    element.preventDefault()
+  props.onclick = function(e) {
+    if (
+      e.button !== 0 ||
```
jbucaran: @SkaterDad What is `e.button`?

src/Link.js
```diff
@@ -4,8 +4,18 @@ export function Link(props, children) {
...
+      e.metaKey ||
+      e.altKey ||
+      e.ctrlKey ||
+      e.shiftKey ||
+      props.target === "_blank"
+    )
```
jbucaran: @SkaterDad Please add `{` and `}`.

src/Link.js
SkaterDad: Want me to rename to `event`?

src/Link.js
SkaterDad: Will do!

test/link.test.js
```diff
@@ -13,6 +13,66 @@ test("/", done => {
...
+
+  const link = h(Link, {
+    to: "foo",
+    go: mock,
+    target: "_blank"
+  })
+
+  link.data.onclick({
+    button: 0, // Left click
+    preventDefault() {} // Noop
```
jbucaran: @SkaterDad Actually `preventDefault` is never called, so maybe let's remove it.

src/Link.js
jbucaran: Sure, let's do that! 🙏

jbucaran:
@SkaterDad Should we care about this?

```
if (window.location.origin !== event.target.origin) {
  return
}
```

jbucaran:
@SkaterDad Other stuff we should pay attention to:

- https://mathiasbynens.github.io/rel-noopener

src/Link.js
```diff
@@ -4,8 +4,19 @@ export function Link(props, children) {
   props.href = props.to
   props.to = null
 
-  props.onclick = function(element) {
-    element.preventDefault()
+  props.onclick = function(event) {
+    if (
+      event.button !== 0 ||
```
jbucaran: @SkaterDad Since we are not (currently) let the user overwrite the `onclick` handler, I think it makes no sense to add [<samp>e.defaultPrevented</samp>](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented) either. What do you think?

SkaterDad: Agreed 👍 

If people have a compelling use case for it, then it's not a very tough addition down the road.

SkaterDad:
@jbucaran The `rel=noopener` stuff is interesting.  I suppose we could add it as a default when people use `target="_blank"` with a link that's on a different origin?  Is that what you were thinking?

Adding the origin check is certainly doable and low cost.  In my own projects, I would probably just use a plain `<a>` tag, so it's more clear to me what is internal, but I can see your point about preferring to use `<link>` for consistency.

Maybe someone else will chime in with an opinion?

jbucaran:
@SkaterDad I was thinking of leaving it entirely up to the user. I prefer doing these things myself because it helps me "learn the platform".

Maybe something like this?

```js
/noopener|noreferrer/.test(event.target.rel)
```

SkaterDad:
Where are you suggesting that code would go?  Mentioning it in the docs
would seem good enough to me since the user needs will all be unique.

Down the road that would make a good candidate for a "development mode"
warning message.

jbucaran:
@SkaterDad Okay, let's think about `noopener` later.

jbucaran:
> @SkaterDad Adding the origin check is certainly doable and low cost.

So, let's do it.

SkaterDad:
@jbucaran Should be good now!  👍 

Had to modify the "test" script to set a URL for the Jest test environment.  That took me way too long to figure out 😆

package.json
```diff
@@ -24,7 +24,7 @@
     "router"
   ],
   "scripts": {
-    "test": "jest --coverage --no-cache",
+    "test": "jest --coverage --no-cache --testURL https://hyperapp.js.org",
```
jbucaran: @SkaterDad Wow, what's this magic?

jbucaran: I see, very nice and very cool find. I certainly didn't know.

Having said that, I'll probably change this to something else, as I don't want the result of the test to depend on a flag specified on package.json

SkaterDad: Yeah, I agree it's a little weird. 

From what I understand, that might be the only way to instantiate a URL in the `jsdom` instances used by `jest`.  f you omit that `testURL`, it defaults to `about:blank`, which makes `window.location.origin === null`.  It throws `SecurityError` if you try to do a `history.pushState(...)` without setting a JSDOM url.

jbucaran: But you should be able to just 

```js
window.location.origin = "https://hyperapp.js.org"
```

before the test.

jbucaran: You will need to do this as well:

```js
Object.defineProperty(window.location, "origin", {
  writable: true
})
```

jbucaran:
@SkaterDad Thanks! 🎉

PR #17, opened 2017-08-26
Link should check event.currentTarget

SkaterDad:
In the <Link> component's origin check, it was checking `event.target.origin`.

While using the component, I quickly discovered it doesn't work correctly if you give the Link any non-text children.

For example:
```js
        Link({to: '/', go: actions.router.go },[
          h('img', {class: 'logo', src: '/static/logo.svg'})
        ]),
```
In this code, when you click the image link, the `event.target` is actually the `img` tag, since you clicked it.  `event.currentTarget` should be used instead, since it always references the element which has the event listener.

Reference docs:  https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

### DIFF
--- a/src/Link.js
+++ b/src/Link.js
@@ -12,7 +12,7 @@ export function Link(props, children) {
       event.ctrlKey ||
       event.shiftKey ||
       props.target === "_blank" ||
-      event.target.origin !== window.location.origin
+      event.currentTarget.origin !== window.location.origin
     ) {
       return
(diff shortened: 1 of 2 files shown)


SkaterDad:
I suppose I should note this behavior is also fixable using CSS, by setting "pointer-events: none" on the link children. 

Not sure if that will sway your opinion on this!

PR #18, opened 2017-08-28
Rename `Router` to `router` to follow lowercase mixin convention.

jbucaran:
- Rename `Router` to `router` to follow lowercase mixin convention.
- Tentative: Make `router` a function call instead: `router()` so we can specify options in the future.
  - But what options? Perhaps { hash: true } as suggested by @jrop?


lukejacksonn:
Yup, I concur on both points. Mixins as a functions makes a lot of sense. Seems much more versatile.

jbucaran:
Thanks guys! 👋

PR #20, opened 2017-09-06
Revamp API: selected <Route /> and <Link /> as "view components".

JorgeBucaran:
## What's this?

Change API as follows:

1) Declare routes by passing them to the `router(..)` mixin, not overloading the view function! 🎉 

```js
mixins: [
  router([ 
    [string, View], 
    [string, View], 
    ... 
  ])
]
```

2) Share matched route with your view as a _view component_. We can define view components as a regular component that is "pre-wired" to your state/actions, etc., and then passed to the view function call as the third argument.

```js
view(state, actions, { Route }) {
  return (
    <main>
      <MyHeader />
        ...
      <Route />
        ...
      <MyFotter />
    </main>
  )
}
```

3) Also share `<Link>` component pre-wired to `actions.router.go` as a view component.

## Example

```jsx
import { h, app } from "hyperapp"
import { router } from "../.."

app({
  view: (state, actions, { Route }) => <Route />,
  mixins: [
    router([
      [
        "/", (state, actions, { Link }) =>
          <Link to="/test">
            Test
          </Link>
      ],
      [
        "/test", (state, actions, { Link }) =>
          <Link to="/">
            Back
          </Link>
      ]
    ])
  ]
})
```

See the [same example](https://github.com/hyperapp/router#hyperapprouter) on the readme to compare.

## Discussion

How do we go about sharing, distributing and "reasoning" about (pre-wired) view components and their simply exported counterparts?

You could still import `Link` from this module and use it as described in the current docs, but if we are going to give it to you pre-wired in the view, why should index.js export Link? What if we don't export Link?

One more thing. What should be the convention for passing "pre-wired" components to the view?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,23 +12,23 @@
 import { router, Link } from "@hyperapp/router"
 
 app({
-  view: [
+  view: (state, actions, { Route }) => <Route />,
+  mixins: [router([
     [
       "/",
-      (state, actions) =>
-        <Link to="/test" go={actions.router.go}>
+      (state, actions, { Link }) =>
+        <Link to="/test">
           Test
         </Link>
     ],
     [
       "/test",
-      (state, actions) =>
-        <Link to="/" go={actions.router.go}>
+      (state, actions, { Link }) =>
+        <Link to="/">
           Back
         </Link>
     ]
-  ],
-  mixins: [router()]
+  ])]
 })
 ```
 
--- a/src/router.js
+++ b/src/router.js
@@ -1,4 +1,6 @@
-export function router(options) {
+import { Link } from "./Link"
+
+export function router(routes) {
   return function(emit) {
     return {
       state: {
@@ -28,13 +30,29 @@ export function router(options) {
           })
         },
         render: function(state, actions, view) {
-          return view[
-            (state.router.index >= 0
-              ? state
-              : actions.router.set(
-                  emit("route", match(location.pathname, view))
-                )).router.index
-          ][1]
+          return function(state, actions, widgets) {
+            widgets = widgets || {}
+
+            widgets.Link = function(props, children) {
+              props.go = actions.router.go
+              return Link(props, children)
+            }
+
+            var route =
+              routes[
+                (state.router.index >= 0
+                  ? state
+                  : actions.router.set(
+                      emit("route", match(location.pathname, routes))
+                    )).router.index
+              ][1]
+
+            widgets.Route = function(props, children) {
+              return route(state, actions, widgets)
+            }
+
+            return view(state, actions, widgets)
+          }
         }
       }
     }


zaceno:
Generally, I like it. But I think the view components should be namespaced to `router`.

You can still destructure like: `(state, actions, {router: {Link, Route}}) => ` or if you want, this is quite nice, valid jsx as well: `<views.router.Route />`

jbucaran:
@zaceno I agree. 

`{router: {Link, Route}}` is the way to go! 👍

lukejacksonn:
Much prefer not having to overload the view function. It is also a good example of a mixin being a function that accepts config. 

Would it be crazy to not have a `Link` at all and instead intercept clicks of any `a` element where the `href` matches a route.. or has that approach been damned as too impure?

zaceno:
@lukejacksonn I think that might interfere with links that are meant to be external (i e don't point to a defined route). I like the idea but I see it getting complicated fast if you're trying to keep it general

SkaterDad:
Another benefit of the `<link>` component is that you can have children inside the `<a>`, and they won't prevent navigation if you click them (since we're checking `event.currentTarget`).  If you do the global `a` handler, you'll need to use CSS `pointer-events: none` to achieve that.

@jbucaran I'm liking the API changes for the most part, but am not really a fan of the array style.  Is there a reason you used that instead of doing a route definition object?  ex:
```js
routes = [
  {
    path: "/",
    view: () => ...
  }
]
```

The objects are more verbose, but allow for future extension, like defining meta data, intial actions, guards, etc.

For most of my use cases, I'd love to be able to also define an action to fire automatically on route changes.  Is that something you feel the router mixin should manage?  Or should I just listen to the `route` event?

zaceno:
@jbucaran, @SkaterDad makes a good point about meta-data & hooks for routes. Authenticated routes is a thing that comes to mind.

jbucaran:
This is how { path, view } would look like:

```js
app({
  view: (state, actions, { Route }) => <Route />,
  mixins: [
    router([
      {
        path: "/",
        view: (state, actions, { Link }) => <Link to="/test">Test</Link>
      },
      {
        path: "/test",
        view: (state, actions, { Link }) => <Link to="/">Back</Link>
      }
    ])
  ]
})
```

Well, I guess it's not bad, but then why not:

```js
{
  routes: [{path,view}, ...],
  other: true,
  options: true
}
```

...and all out of a sudden it looks complicated. Or maybe that's just me too used to the old router?

jbucaran:
> @lukejacksonn Much prefer not having to overload the view function. 

Then how do we pass the matched route function to the view?

zaceno:
While I do like the object-method of defining routes, I think we could discuss it in a separate, future issue/PR. Because when we eventually need it, it could easily be implemented in a backwards compatible way. (`[x, y] => {path: x, view: y}`)

lukejacksonn:
> I think that might interfere with links that are meant to be external

@zaceno the old router checked if the href started with a `/` in order to decide wether to intercept or not.. this could probably be improved on further and only intercept if a route for the path was defined. But I do appreciate this is approach is much harder to reason about than a `Link` component.

As @jbucaran and @SkaterDad point out some advantages of `Link`:

- A component is an improvement on intercepting and processing all window clicks
- You can have children inside the component and they won't prevent navigation

> The idea of passing components to the view might actually be a phenomenal idea!

I agree! Let's keep exploring it. This might not be the best use case to show it's full potential.

Seems like we are tending toward view decorator functions (not sure if it the right terminology but it sounds similar; by definition, a decorator is a function that takes another function and extends the behavior of the latter function without explicitly modifying it)?

jbucaran:
@lukejacksonn Yes, by definition you are completely right. We are "wrapping" the view function into another function and using that instead to produce a vtree.

zaceno:
>> The idea of passing components to the view might actually be a phenomenal idea!
> I agree! Let's keep exploring it. This might not be the best use case to show it's full potential.

I like the idea too. In fact I think once the router establishes it as the standard, other mixins will want to follow. And I think we need a strategy here.

The "way" you do it is to decorate the view in the render event. Once multiple mixins are doing this, there will be a cascade of render events and decorations of the view in order for each mixin to "hang" it's view components on the tree.

Maybe that's ok. Or maybe hyperapp could allow you to define view components in mixins. Not sure which is better. But it's worth considering. Perhaps in a separate issue/PR though, that too.

README.md
```diff
@@ -12,23 +12,23 @@
...
+      (state, actions, { Link }) =>
+        <Link to="/test">
           Test
         </Link>
     ],
     [
       "/test",
-      (state, actions) =>
-        <Link to="/" go={actions.router.go}>
+      (state, actions, { Link }) =>
```
zaceno: `{router: {Link}}`

src/router.js
```diff
@@ -28,13 +30,29 @@ export function router(options) {
...
-            (state.router.index >= 0
-              ? state
-              : actions.router.set(
-                  emit("route", match(location.pathname, view))
-                )).router.index
-          ][1]
+          return function(state, actions, widgets) {
+            widgets = widgets || {}
+
+            widgets.Link = function(props, children) {
```
zaceno: Attach `Route` and `Link` to `widgets.router`, not just `widgets`

jbucaran:
@zaceno How would it work? You know, mixins taking views. And how could mixins take more than one view without mixins in mixins?

zaceno:
@jbucaran -- I didn't mean mixins should (by default) be *consuming* views (that would be mixins in mixins). I meant that mixins will want to define and export multiple views (ahem *view components*). Just like you're proposing here with the router (which defines `Link` and `Router`

jbucaran:
@zaceno Ah, yes, exactly. So, what could be the use for `view` in mixins? Just brainstorming here.

zaceno:
> So, what could be the use for view in mixins? Just brainstorming here.

Good question. I don't think `view` (singular) has any obvious use in a mixin, since I imagine quite a few will want to define more than one view components. I think `views` or `components` (plural) would be more appropriate.

But that introduces an asymmetry... I know...

andyrj:
I agree with @SkaterDad, I prefer the [{path, view, meta...}] to [[path, view, meta...]], it's what I've been doing for my routing stuff.

API change looks good to me.  I like this approach for changing away from the overloaded app.view, this will make a lot of people happy that have been asking about how to have a static layout that only changes a portion of the "view" during route change...

MatejMazur:
You can avoid the third argument via closures:

```
var closure = (function() {
  var emit;

  function createMixin(model) {
    return function(options) {
      return function(_emit) {
        emit = _emit
        var mixin = model(options)(emit)
        mixin.events.getStateActions = function(state, actions) {
          return {
			  state: state, 
			  actions: actions
		  };
        }
        return mixin
      }
    }
  }

  function createWidget(widget) {
    return function (props, children) {
	  var sync = emit('getStateActions')
      return widget(sync.state, sync.actions)(props, children);
    }
  }

  return {
    createMixin,
    createWidget
  };
})()
```

Now you just create `router` and `Link` like this:

```
var router = closure.createMixin(function(options) {...})
var Link = closure.createWidget(function(state, actions) {
  return function(props, children) {...}
})
```

Here is working example of the router with `closure` function: https://codepen.io/MatejMazur/pen/qXGQWX?editors=0010

And here is test of this router:
https://codepen.io/MatejMazur/pen/JyqeKr?editors=0010

Just my 2 cents 😄

zaceno:
Personally, I'm a fan of passing pre-wired components to the view, because I'm already using that pattern via hyperapp-partial.

For other's however, who perhaops use Link and Route somewhere deep in a tree of components, they are passing `state` and `actions` as props down the chain.

This change means that they will have to pass the third argument down the chain instead of -- or possibly *in addition to* -- `state` and `actions`.

So for them, it's an extra, wierd thing they need to consider, just because they used the router. 

Now I imagine/hope the discomfort will be temporary, as more and more mixins and components are made this way. But still... I understand the hesitance.

jbucaran:
> @zaceno This change means that they will have to pass the third argument down the chain instead of -- or possibly in addition to -- state and actions.

I don't understand what you mean. Can you elaborate on this? 🤔

zaceno:
@jbucaran: I haven't got a specific example in mind. But, say someone was doing something like this:

```
import {router, Link} from 'hyperapp-router'

const Bar = ({state, actions}) => <div>
   {state.something && <button onclick={actions.doSomething}>Click me</button>
   <Link to="/elsewhere" go={actions.router.go}>Go elsewhere</Link>
</div>

const Foo = ({state, actions}) => <div>
  Here's a Bar:
  <Bar state={state} actions={actions} />
</div>

const Main = (state, actions) => <div><Foo state={state} actions={actions} /></div>

app({
  mixins: [router]
  state: {...}
  actions: {...},
  view: [["/": Main], ["/elsewhere", ..]]
})

```


with this change, that would have to be written: 

```
import {router, Link} from 'hyperapp-router'

const Bar = ({state, actions, views}) => <div>
   {state.something && <button onclick={actions.doSomething}>Click me</button>
   <views.router.Link to="/elsewhere">Go elsewhere</Link>
</div>

const Foo = ({state, actions}) => <div>
  Here's a Bar:
  <Bar state={state} actions={actions} views={views} />
</div>

const Main = (state, actions, views) => <div><Foo state={state} actions={actions} views={views} /></div>

app({
  mixins: [router([
    ["/": Main],
   ["/elsewhere", ...],
  ])],
  state: {...}
  actions: {...},
  view: (state, actions, views) => views.router.Route()
})

```

In other word: more verbose and not obviously helpful. It get's more useful when also Foo and Bar can be prewired.

jbucaran:
@zaceno I see what you are saying now. It's not that "with this change" we are forcing you to write anything in a particular way. 

You can still write things the old way if you like, but you'd need to wire Link to `actions.router.go` yourself.

zaceno:
@jbucaran ah so there will be two ways to use/access link? The non-prewired component via import, and the prewired one via 3d arg? I like the sound of that 👍

jbucaran:
If we export it, yes. But seriously, I am starting to have second thoughts about passing stuff to the view and I feel we need to think more about it.

The problem I have with it is that there seem to be two kind of components, at least. Components like the Link can be repeated in many places across your UI, but there is only one location.pathname, so clicking in one link does not affect other links.

But what about collection-like components that would be represented in the state as an array of objects of some structure, like a list of counters.

zaceno:
@jbucaran I don't see a problem with collections of things. The way I envision it, the counter view component would be used by passing an id/key in the props, representing the counter in the state to show. Being pre wired to the state, the view component then knows how to look up the specific counter data.

zaceno:
It's a simplification to only have to pass an Id to the component, rather than have to type the path to the specific value, and actions with bound arguments.

zaceno:
It's not everyone's cup of tea for sure, it's a trade off of explicitness and brevity. Depends what you find more readable.

jbucaran:
We need examples of that! :)

zaceno:
@jbucaran here's a very simple example of what I meant (based on partials) https://codepen.io/zaceno/pen/xLvzGB?editors=0010

zaceno:
Actually, @MatejMazur 's idea of how to export pre-wired components (instead of providing via third argument) is really cool, but perhaps not explained as simply as it could be. Here's my version of the same idea:

```

//A mixin that can make prewired view components importable/exportable

var prewired = (function() {
  var tmp;
  
  var mixin = function (emit) {
    return {
      events: {
        render: function (state, actions, view) {
          tmp = [state, actions]
          return view
        }
      }
    }
  }

  function createMixin(model) {
    return function(options) {
      return function(emit) {
        var mixin = model(options)(emit);
        mixin.events.render.push(function(state, actions, view) {
          stack.push({
            state: state,
            actions: actions
          });
          var result = view(state, actions);
          stack.pop();
          return function() {
            return result;
          };
        });
        return mixin;
      };
    };
  }
  
  mixin.view = function (f) {
    return function (props, children) {
      return f(tmp[0], tmp[1], props, children)
    }
  }

  return mixin
})()
```

Now, when your app uses this mixin (`app({ mixins: [prewired, ...]}`, you can define components like this:

```
export const Link = prewired.view((state, actions, props, children) =>
  <a href="." onclick={_ => actions.router.go(props.to)}>{children}</a>
)

```

So for the router to use this, it'd have to bake in the prewired mixin described above. 

It's a cool idea, worth considering!

Still, I'm not 100% on board with this. I mean: it works, but something about it bugs me. Somehow, passing prewired view components via arguments feels more "right" since we're already getting the state and actions that way. This way seems more indirect and magical.

zaceno:
@jbucaran Same example as before (of collection of prewired components). This time using @MatejMazur 's idea, and so providing the view components via module scope rather than 3d argument.

https://codepen.io/zaceno/pen/LjwBJO?editors=0010

SkaterDad:
These ideas about pre-wired view components are really creative!

Passing a 3rd argument to the `view` function is probably my favorite of these.  It's essentially mimicking the state & actions if people are good with namespacing, so it seems more `hyperapp`.

jbucaran:
@zaceno We are on the right track, but some things are still not crystal clear to me.

For example, Link is a simple component. It's prewired to actions.router.go, but it still accepts other props. You can even overwrite the go action via props.go.

The selected Route, on the other hand is not a component, but a view. You just put it there `<Route /> and it takes no props. If it did, what could those props be?

Thanks Zaceno for your example using indices with hyperapp-partial, I still need to tinker a little bit more with it to see how things would look like without any external libraries.

zaceno:
@jbucaran Technically, the `<views.router.Route />` *is* a component (the way I define it anyway ;) ). I mean: it is a function that returns a vtree. The vtree it returns just happens to come from calling the routed view. 

You *could* pass props & children to it (`<views.router.Route aprop="foo" />`) -- they just don't have any effect. Which is the same as passing any irrelevant prop to any other component.

So I don't really see a problem with `Route` except perhaps the name. Maybe it should be called `RoutedView`, `MatchedView` or perhaps (since we've got the `router` namespace) `<views.router.matched />`, `<views.router.routed />` or `<views.router.view />`

SkaterDad:
@jbucaran Theoretically you could merge any props passed to `<views.router.Route />` into the router view's root node.  It could be used to apply a common class name to all of the page containers, for example.

okwolf:
@JorgeBucaran should we close this in favor of #23?

JorgeBucaran:
Yes. Thanks @okwolf!

PR #22, opened 2017-10-08
hoa router

speier:
router implemented as a higher order app (hoa)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,30 +6,44 @@
 
 @hyperapp/router provides utilities for routing client-side pages with [Hyperapp](https://github.com/hyperapp/hyperapp) using the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History).
 
+
+Use the router as a Higher Order App [HOA](https://github.com/hyperapp/hyperapp/releases#higher-order-apps). Then compose your view as an array of [routes](#routes).
+
+
 [Try it Online](http://hyperapp-router.surge.sh)
 
 ```jsx
 import { router, Link } from "@hyperapp/router"
 
-app({
-  view: [
-    [
-      "/",
-      (state, actions) =>
-        <Link to="/test" go={actions.router.go}>
-          Test
-        </Link>
-    ],
-    [
-      "/test",
-      (state, actions) =>
-        <Link to="/" go={actions.router.go}>
-          Back
-        </Link>
-    ]
+const routes = [
+  [
+    "/",
+    (state, actions) =>
+      <Link to="/test" go={actions.router.go}>
+        Test
+      </Link>
   ],
-  mixins: [router()]
+  [
+    "/test",
+    (state, actions) =>
+      <Link to="/" go={actions.router.go}>
+        Back
+      </Link>
+  ]
+]
+
+const routable = router(routes)
+
+app(routable)({
+  state: {
+    count: 0
+  },
+  actions: {
+    down: state => ({ count: state.count - 1 }),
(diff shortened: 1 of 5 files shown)


src/router.js
```diff
@@ -1,41 +1,38 @@
...
+              actions.set({ path: path })
             }
           }
         }
-      },
-      events: {
-        load: function(state, actions) {
-          addEventListener("popstate", function() {
+
+        props.view = function (state, actions) {
```
pspeter3: This seems like a potentially dangerous pattern with HOA. What happens if a view was defined in the existing app chain?

src/router.js
```diff
@@ -1,41 +1,38 @@
...
+          window.onpopstate = function () {
```
pspeter3: Can this be an event listener as opposed to a handler? Overriding global state again seems concerning. Also why does this happen in view as opposed to looking at the actions returned by the call to app?

speier: unfortunately it wasn't working outside of the view. i prefer to use `addEventListener` outside the view but can't make it working.

src/router.js
speier: events gone, so the only possibly entrypoint is the view function, but i think that is ok. the previous implementation of the router was doing pretty much the same on the `render` event.

speier: please see my lates commit, that is overwriting the default view only when the route matches

src/router.js
speier: please see may improved implementation in the latest commit; `addEventListener` working

src/router.js
```diff
@@ -1,41 +1,41 @@
...
+      const actions = app(enhance(props))
+
+      window.addEventListener('popstate', function () {
+        actions.router.set({})
+      })
+
+      return actions
+
+      function enhance(props) {
+        props = Object.assign({ state: {}, actions: {} }, props)
```
SkaterDad: hyperapp's core aims to support IE10+, so you may need to replace `Object.assign` with a custom function to maintain that compatibility.

Maybe @JorgeBucaran would consider exporting the `merge()` function from hyperapp/src/app.js so other packages can use it and avoid duplicating efforts/code?

https://github.com/hyperapp/hyperapp/blob/master/src/app.js#L120-L129

It's also possible that his latest PR to introduce modules will mean you don't need this section of code.

speier: ok, i will fix that. exporting `merge` would be great. in the meantime how about simply?
```
props = props || {}
props.state = props.state || {}
props.actions = props.actions || {}
```

also, i will check out modules...

Swizz:
Using module we do not need HOA anymore :tada:

README.md
```diff
@@ -6,30 +6,44 @@
...
+    (state, actions) =>
+      <Link to="/" go={actions.router.go}>
+        Back
+      </Link>
+  ]
+]
+
+const routable = router(routes)
+
+app(routable)({
```
okwolf: Support for passing HOAs to `app` has been removed in the latest Hyperapp

package.json
```diff
@@ -37,13 +37,13 @@
   },
   "devDependencies": {
     "babel-preset-es2015": "^6.24.1",
-    "hyperapp": "^0.12.0",
-    "jest": "^20.0.4",
-    "prettier": "~1.5.3",
-    "rollup": "^0.47.6",
+    "hyperapp": "^0.14.0",
```
okwolf: Hyperapp 0.15.1 is out now

PR #23, opened 2017-10-19
Component Router

speier:
Basic component router implementation.

```jsx
import { h, app } from 'hyperapp';
import { router, Router, Route } from '@hyperapp/router';

app({
  state: { ... },
  actions: { ... },
  view: (state, actions) => (
    <Router state={state} actions={actions}>
      <Route path="/" view={Home} />
      <Route path="/login" view={Login} />
      <Route path="/404" view={NotFound} />
    </Router>
  ),
  modules: { router }
})
```


src/router.js
```diff
@@ -1,72 +1,64 @@
...
+    },
+    go: function (state, actions, path) {
+      if (location.pathname + location.search !== path) {
+        history.pushState({}, '', path)
+        actions.set({ path: path })
       }
     }
+  },
+  init(state, actions) {
+    window.addEventListener('popstate', function () {
```
JorgeBucaran: @speier Pretty sure you can drop the `window` here.

README.md
```diff
@@ -6,29 +6,27 @@
...
-        <Link to="/" go={actions.router.go}>
-          Back
-        </Link>
-    ]
-  ],
-  mixins: [router()]
+  state: { ... },
+  actions: { ... },
+  view: (state, actions) => (
+    <Router state={state} actions={actions}>
```
JorgeBucaran: @speier So, it's decided we are going to share both state and actions with each component/view? 🤔👍

README.md
```diff
@@ -6,29 +6,27 @@
 
 @hyperapp/router provides utilities for routing client-side pages with [Hyperapp](https://github.com/hyperapp/hyperapp) using the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History).
 
+
+Use the `router` as Hyperapp module. Then compose your routes with components (`Router`, `Route`).
```
JorgeBucaran: Maybe:

> Use `router` as a Hyperapp module.

/cc @zaceno @lukejacksonn

README.md
```diff
@@ -58,20 +56,6 @@ Then [bundle](https://github.com/hyperapp/hyperapp/blob/master/docs/getting-star
...
-app({
-  view: [
-    ["/", state => <h1>Hi.</h1>]
-    ["*", state => <h1>404</h1>],
-  ],
-  mixins: [router()]
-})
-```
-
 ### Routes
```
JorgeBucaran: @speier Maybe remove all this outdated stuff?

README.md
zaceno: Yes "... as *a* module ...".

src/router.js
```diff
@@ -1,72 +1,64 @@
...
+      actions.set({})
+    })
   }
 }
 
-function match(pathname, routes) {
+// Router component
+export function Router(props, children) {
+  const m = match(location.pathname, children)
```
JorgeBucaran: @speier Could you please use `match` instead of `m`, `index` instead of `i` and `view` instead of `v`. 🙏

src/router.js
```diff
@@ -1,72 +1,64 @@
...
-            (state.router.index >= 0
-              ? state
-              : actions.router.set(
-                  emit("route", match(location.pathname, view))
-                )).router.index
-          ][1]
-        }
+import { h } from 'hyperapp'
+
+// router module
```
JorgeBucaran: @speier Could you please move `router`, `Router` and `Route` into their own files:

- `router.js`
- `Router.js`
- `Route.js`

On-topic: Router & Route is familiar coming from React, but what  alternatives do we have? Just curious.

src/router.js
```diff
@@ -1,72 +1,64 @@
...
+  },
+  init(state, actions) {
+    window.addEventListener('popstate', function () {
+      actions.set({})
+    })
   }
 }
 
-function match(pathname, routes) {
+// Router component
```
JorgeBucaran: Router.js

src/router.js
```diff
@@ -1,72 +1,64 @@
...
+export function Router(props, children) {
+  const m = match(location.pathname, children)
+  const i = m.match ? m.index : children.length - 1
+  const v = children[i].component
+  return v(props.state, props.actions)
+}
+
+// Route component
```
JorgeBucaran: Route.js

src/router.js
```diff
@@ -1,72 +1,64 @@
...
+export function Route(props, children) {
+  return props
```
JorgeBucaran: @speier So much room for activities here? :) 

If you are not using `children`, please remove it.

src/router.js
```diff
@@ -1,72 +1,64 @@
...
       RegExp(
         route === "*"
           ? ".*"
           : "^" +
-            route.replace(/\//g, "\\/").replace(/:([\w]+)/g, function(_, key) {
-              keys.push(key)
-              return "([-\\.%\\w\\(\\)]+)"
-            }) +
-            "/?$",
+          route.replace(/\//g, "\\/").replace(/:([\w]+)/g, function (_, key) {
```
JorgeBucaran: @speier @lukejacksonn I know it deserves its own proper issue, but what about parsing the query string too? Just saying.

README.md
lukejacksonn: > Include `router` as a module in your app.

Sounds good

src/router.js
lukejacksonn: We had this discussion before.. it was dismissed on the grounds that whatever implementation we chose would not be sufficient enough (for all cases).. that said, I still think that a simple query string parser would be useful for most people.

JorgeBucaran: Thanks @lukejacksonn, so it's just a matter of deciding how simple is simple enough and lay it out for later implementation. Let's discuss this later then.

speier:
Thanks for the feedback! I'm looking into the comments and push changes soon.

src/router.js
speier: ✅

README.md
zaceno: It's either:
A:
```jsx
MyComponent = props => <div>{props.foo}</div>

app({
  modules: {router},
  state: {foo: 'bar'},
  view: (state, actions) => (
    <Router state={state} actions={actions}>
      <Route path="/foo" component={_ => <MyComponent foo={state.foo} />} />
    </Router>
  )
})
```

B:

```jsx
MyComponent = props => <div>{props.state.foo}</div>

app({
  modules: {router},
  state: {foo: 'bar'},
  view: (state, actions) => (
    <Router state={state} actions={actions}>
      <Route path="/foo" component=MyComponent />
    </Router>
  )
})
```

or C:


```jsx
MyComponent = (state, actions) => <div>{state.foo}</div>

app({
  modules: {router},
  state: {foo: 'bar'},
  view: (state, actions) => (
    <Router state={state} actions={actions}>
      <Route path="/foo" component=MyComponent />
    </Router>
  )
})
```

I prefer option *B*

src/router.js
speier: ✅

README.md
speier: 👍

src/router.js
speier: 👍

src/router.js
speier: 👍

src/router.js
speier: 👍

README.md
JorgeBucaran: @speier @zaceno I think I get it now. These are not components, React calls it components, since that's the terminology they have, but what we are dealing with here are just views, just like the old router used to route views, not components. 

I think the way to move forward here is to just rename `component` to `view` and go with what Kalman made.

src/router.js
speier: ok, separated to their respective files.

Yes, Router & Route are coming from the react ecosystem probably. Not sure about any valid alternatives.

README.md
speier: 👍

README.md
zaceno: ...and *D* is just *B* with a better name. So I'm good with that 😄 👍

speier: great, let's go with _B_ using the name __view__, so +1 for _D_ ;)

src/router.js
lukejacksonn: Yep, exactly! I just tried finding the thread but can't remember where it was discussed 🤔 There was talk of doing what react router does, which IIRC, allows the developer to pass in a query string parsing function; that way we can't be at fault. Perhaps allowing for that, but providing a default parser suitable for >90% of use cases is the way to go.

JorgeBucaran:
@speier Can you please create `router.js` for the router module and use `index.js` just to export all the stuff?

speier:
@JorgeBucaran tbh, i don't really like the idea of having two files with the same name (but different case `router.js` vs `Router.js`)

JorgeBucaran:
@speier We don't really have a choice. 😆

speier:
@JorgeBucaran actually I can't do that even I would, because I'm on APFS (case-sensitive), not to mention Windows users. :)

JorgeBucaran:
Wham. So, what options do we have? Clearly APFS and Windows never imagined someday we would rain on their parade. 😆

zaceno:
```
src/
    index.js
    router-module.js
    router-component.js
    route-component.js
    link-component.js
```
?

lukejacksonn:
We could do something like..

@hyperapp/router
|--- index.js
|--- router.js
|--- views
      |--- router.js
      |--- route.js

zaceno:
Btw is there no `Link` component? Not that I miss it much -- it would mostly be valuable if it were pre-bound to `actions.router.go` but anyway: what's the story there?

speier:
how about this ?
```
src
├── components
│  ├── Link.js
│  ├── Route.js
│  └── Router.js
├── index.js
└── router.js

```

zaceno:
@speier looks good to me. Only: if we're not expecting users' filesystems to be able to differentiate between upper-/lower-case, then we might as well standardize on lowercase for filenames (even for components). Just my pair-o-pennies

speier:
@zaceno sure, i was about thinking the same.. filenames will be lowercase. 👍

SkaterDad:
@JorgeBucaran @speier @lukejacksonn 
These changes look pretty good, but I fear this version of the router has a lot less power than before.

The removal of custom events & isolation of modules has left me very confused about how I'm going to upgrade my app to the latest `hyperapp`.   Can we have some discussion on how to accomplish these common routing related tasks?

- Route entry guarding (ex:  check if user is authenticated before allowing route change)
  - This needs access to the full state tree
- Data fetching on route change
  - This could need either the global actions, or module actions.

I've done a lot of experimenting so far (with an HOA, and also a super-powered Link onclick function), but I haven't found a solution I'm happy with.

zaceno:
@SkaterDad I would also like to have a deeper discussion about the points you mention! 👍  But, if those problems arise out of module isolation and lack of event system, then I wouldn't say it's the router itself that is less powerful -- rather core hyperapp itself.  Perhaps open issues for those points in the main hyperapp repo and continue there?

JorgeBucaran:
@SkaterDad 

> Data fetching on route change

`onroute` is not implemented in this PR, but it should be available on a later PR or @speier can give it a shot if he wants.

> Route entry guarding (ex: check if user is authenticated before allowing route change)

Sounds like something `onroute` could help with.

---

@speier 

```
src
├── components
│  ├── Link.js
│  ├── Route.js
│  └── Router.js
├── index.js
└── router.js
```

LGTM

zaceno:
@JorgeBucaran Just for clarification: when you say `onroute`  you're talking about a prop on the `Route` component right? (I had to think about it a sec before I think I understood what you meant.)

JorgeBucaran:
@zaceno Yep, a prop in the Router or Route component. Then using the other actions receives via prop setup the infrastructure for handling the route change.

zaceno:
@JorgeBucaran Right, thanks! @SkaterDad so maybe this is the right place (right repo at least) to bring up those questions after all :)

SkaterDad:
I think you need 2 separate hooks:
`beforeroute`/`onenter` - logic which can cancel the navigation or redirect before the URL is changed.
`onroute` - things to do after the navigation was applied

Can you shed some light on where you think that logic will live?

I've played with the idea that the router should *be* the application, and would therefore have access to the full state/actions.  I think it's how `elm`'s navigation program works.

speier:
@JorgeBucaran component names are UpperCase (`Router`, `Route`), but I would rather keep the filenames lowercase for consistency. Is that a problem?

lukejacksonn:
I would tend to agree with that.. uppercase file names are obscure. The only reason the notion of capitalising components comes from JSX and we don't all have to abide by those norms. They are clearly components.. they are nested inside a `components` directory. I don't see any justification for capitalising other than personal preference but perhaps I am missing something.

JorgeBucaran:
@speier React is not my "moral compass", but it's a reasonable convention they have.

> I would rather keep the filenames lowercase for consistency.

For consistency with other file names, that is. It is, however, inconsistent with the name of the component in the code, which is capitalized.

**Having said that..:**

> They are clearly components.. they are nested inside a components directory.

Good point! If components are **always** under `components/` there is certainly no ambiguity. 

What does everyone else think? Your feedback and comments please. 🙏 

/cc @rajaraodv @selfup @Swizz @okwolf @andyrj @johanalkstal @jrop @chiefsmurph

JorgeBucaran:
After some research I found a reputable source that would encourage `Router.js` instead of `router.js`.

https://github.com/airbnb/javascript/#naming--filename-matches-export

> 23.6 A base filename should exactly match the name of its default export.
> 23.7 Use camelCase when you export-default a function. Your filename should be identical to your function’s name.
> 23.8 Use PascalCase when you export a constructor / class / singleton / function library / bare object.

speier:
ok, i'm perfectly fine with that approach as well.

actually my only reason against the non-lowercase approach is historical, i've seen issues when devs tried to require modules with different names, which was working fine on mac/win laptops, but caused errors on linux environments (luckily intercepted on staging first).

okwolf:
I’m used to following PascalCase for component files from React, so I’m fine with this convention for HA.

src/components/router.js
```diff
@@ -0,0 +1,47 @@
+export function Router(props, children) {
+  const match = props.actions.router.set(matcher(location.pathname, children))
```
SkaterDad: Should this action be called within the router module instead?  Calling it with `location.pathname` makes the view an impure function.

This was only necessary because the router module's init method is calling `actions.set({})`.  Why not call `go` instead?
```js
// In router module

    init(state, actions) {
      actions.go(window.location.pathname + window.location.search)
      addEventListener("popstate", function() {
        actions.go(window.location.pathname + window.location.search)
      })
    },
```

While we're at it, should `matcher` be called in the `go` action also?

JorgeBucaran: @speier Ping.

selfup:
Yea let's make it official to match filenames with the export name as the airbnb rules said. 🎉 

That way, it's simple.

I honestly don't really mind snake case or kebab case filenames but it would be nice to not have to worry about this coming up again 😂

vdsabev:
I ❤️ where you're going with the component router!

Here are some use cases I have:
1. Show transition animations when removing the old route's content and showing the new one. Ideally, I should be able to only define the transitions once in the `Router`, e.g.:
```jsx
  <Router
    onroutecreate={/* some code to fade new content in */}
    onrouteremove={/* some code to fade old content out */}
  >
    <Route />
    <Route />
  </Router>
```
2. Asynchronously loading data **before** the route is navigated to. Perhaps it could be within a single event that is **cancellable**, either via promise or callbacks. One example would be authenticating the user asynchronously to check whether they have access to the route; if they are, load relevant data for the view and finally resolve, otherwise - cancel route navigation or throw error to be handled upstream
3. Show a Loader view while the router is resolving data asynchronously
4. Setting the title of the page, usually after all data has resolved, e.g. when I navigate from `/home` to `/profiles/me`, I want to change the title from `My Website - Home` to `My Website - Vladimir Sabev`
5. Router within router? E.g. if I have a home page at `/` and I want to have `/login` and `/register` within that page, I could do something like this:
```jsx
  <Router>
    <Route path="/" view={Home} />
  </Router>
```
```jsx
const Home = () =>
  <Router>
    <Route path="/login" view={Login} />
    <Route path="/register" view={Register} />
  </Router>
```

I have some of this functionality in my own implementation of the router:
1. See [app.js](https://github.com/vdsabev/civvy/blob/2d4f196a8babcf7bd191f6941021d3beb79e16f5/src/app.js) for usage
2. [Page.js](https://github.com/vdsabev/civvy/blob/2d4f196a8babcf7bd191f6941021d3beb79e16f5/src/page/Page.js) - this is supposed to be like the `Router` and `Route` components
3. [router.js](https://github.com/vdsabev/civvy/blob/2d4f196a8babcf7bd191f6941021d3beb79e16f5/src/router.js)

zaceno:
@vdsabev Except for the nested-routing use case, all the others should be solvable I think.

For transitions, you could use https://github.com/zaceno/hyperapp-transitions  (or roll your own of course) You wouldn't be able to set it as props on the Router component like you described, but this should work:

```jsx

const pageTransition = ... // define transition

const Page = (props,content) => pageTransition(h('div', {class: "page", key: props.key}, content))

const AboutPage = props => Page({key: 'about'}, [...])

const BlogPage = props => Page({key: 'blog'}, [...])
...

app({

view: (state, actions) => <Router>
  <Route path="/about" view={AboutPage} />
  <Route path="/blog" view={BlogPage} />
  ...
</Router>

})
```

Swizz:
This is just a simple question but why did you dissmissed the slot/children solution for wiring component to a route ?

```jsx
<Router state={state} actions={actions}>
  <Route path="/foo">
    <MyComponent/>
  </Route>
  <Route path="*">
    <h1>404 Not Found</h1>
    <p>Unicorn does not exist</p>
  </Route>
</Router>
```

zaceno:
@Swizz The reason we can't do that, is because that would evaluate *all* the routes children and build their vtrees *before* matching the route. That would work for very small apps but would not scale at all.

You could do this, on the other hand:

```jsx
<Route path="*" view={_ => <div>
  <h1>404 Not Found</h1>
  <p>Unicorn does not exist</p>
</div>} />
```

lukejacksonn:
@swizz I agree.. this option should be explored and considered. I have found a very simple solution to the problem @zaceno is proposing. This approach also makes nested routing trivial.

```js

const Route = ({ exact, path = '', view }) => {
  const match = exact
    ? location.pathname === path
    : location.pathname.match(path)
  return match && view({ path })
}

const Switch = children =>
  children.reduce((a, b) => (b && !a.length ? a.concat(b) : a), [])


app(
  view: (state, actions) =>
    Switch([
      Route({ path: '/', view: Home(state, actions), exact: true }),
      Route({ path: '/auth', view: Login(state, actions) }),
      Route({ path: '/profile', view: Profile(state, actions) }),
      Route({ view: Lost(state, actions) }),
    ])
)
```

The last route will only match if none of the routes before it match.

zaceno:
@lukejacksonn ...I don't get it. How is `Home(state, actions)` *not* prematurely evaluated in your example?

lukejacksonn:
Because route only returns a view if the path matches, otherwise it returns false.
```
match && view({ ... })
```
Switch will return the first truthy route (child) only.

lukejacksonn:
This is a very basic `Route` for example.. in the real implementation it also plucks out route and query string params and passes them on to the Route view as props along with path.

zaceno:
Ah so Home(..) IS evaluated, but it only returns another function - not a vtree. Well that's cool then :) 

But it's not was @Swizz was after, I think. If I'm not mistaken, his thing is he wants to be able to define the matched view as children of the matching route component.

lukejacksonn:
Well in that case.. you can just have `Home` etc. return an array instead of a component!
```js
const HomeChildren = (state, actions) => (props) => [
  h1('Unwrapped Children'),
  p('for my matching route')
]
```

Swizz:
You are right about the Vtree, I was mistaken about that.
I what just founding this solution more neat than `<Route view={Component}/>`.

But now I understand why.

zaceno:
@Swizz I agree -- that would have been nice! Explored it a while back and that's how I first encountered the problem.

Swizz:
And what about the following?

```jsx
<Router state={state} actions={actions}>
  <Route path="/foo" namespace="foo">
    { (state, actions) => <MyComponent state={state} actions={actions}/> }
  </Route>
  <Route path="*" namespace="notfound">
    { (state, actions) => 
      <div>
        <h1>404 Not Found</h1>
        <p>Unicorn does not exist</p>
      </div>
    }
  </Route>
</Router>
```

We can create a convention like, in hyperapp : 
* You can use a **component** : A function that return a vdom tree
* You can use a **slot** : A function that return a function that use the state and action to return a vdom tree

**slot** will be useful to create component that use namespaced state/actions and to create the vdom tree only on demand.

zaceno:
@Swizz, with that syntax, we are passing the views inside a `children` array. So when the route matches, should the router  `return children.map(view => view(state, actions))` ?  Does having multiple children like that make sense? (Maybe it does, I don't know).

But if we decide that multiple children (in the form of view-functions) *doesn't* make sense, I think the syntax had better be:

```
<Route path="/foo" view={(state, actions) => (
  <div>
    <h1>Behold the Foo!</h1>
    <p> My foobar brings all the boys to the yard... </p>
  </div>
)} />
```

Swizz:
What is happening if I define two routes with the same path in my `Router` ?
What is a `Router` conceptually ?

On my side I am only seeing a "visibility" check.
> If I am on the "foo" route add to the vtree this node
So I don't see any trouble to use `Router` at multiple place in the same Vtree. But what is your thought about this everybody ?

lukejacksonn:
@Swizz in my proposed router there is no such component as `Router` as there is no need (see my example with `Route` and `Switch` above). All a `Route` does is what you say.. If I am on the "foo" route add to the vtree this node.

> What is happening if I define two routes with the same path in my Router ?

In my model they will both show up (unless you use `Switch` which will return the first match). You can nest routes  as much as you want and create routes dynamically. For example:

```
const Friends = (state, actions) => ({ path }) =>
  ul(state.friends.map(li))

const Profile = (state, actions) => ({ path }) => [
  !state.user && Redirect({ to: '/login' }),
  h1(`Hello ${state.user}`),
  Link({ to: `${path}/friends` }),
  Route({ path: `${path}/friends`, view: Friends(state.user, actions.user) })
]

view: (state, actions) => main([
  Route({ path: '/', exact: true, view: Home }),
  Route({ path: '/profile', view: ({ path }) => Profile(state, actions) })
])
```

Passing the state/actions like this feels kind of like _prewiring_ the _view_ with what it needs, turning it into a _component_ that takes props.

JorgeBucaran:
Thank you @speier for all your work on this PR! I am going to close now in favor of #28.

PR #26, opened 2017-11-06
Another take on a component router

lukejacksonn:
I was inspired by [react-router](https://github.com/ReactTraining/react-router) to create something similar (but simpler) for hyperapps. The main difference between this router and the old router is that there is very little state/actions here, it is almost entirely Component based.

Highlight features include:

- Dynamic nested routes.
- No need to pass `actions.router.go` to Link.
- Route guards using Redirects.
- Shiny new [documentation](https://github.com/lukejacksonn/router/blob/941276f0acff7be8f103a3598e04c94731503e68/README.md).
- Route params _and_ queryString parsing.

It works! No tests yet though 🙃 not a complete PR but looking for feedback 🙇


SkaterDad:
Looking good so far.

Do you have an example prepared for data fetching?

lukejacksonn:
@SkaterDad I know what you mean about the `Popstate` component, there was a big thread on slack today about it and for now (since the purge of init) we think it is our best option. The alternative is adding `addEventListener('popstate', actions.router.update)` after your app that returns `actions`. Technically you can do either, but having everything as a component feels more uniform and easier to explain.

So route views are just functions that return a vnode, so you can wire then with whatever state/actions you like. Personally I would curry them (like actions):

```
const Home = (state, actions) => ({ path, params, query }) => h1(state.message)
const Away = ({ path, params, query }) => h1('Bye')

app({
  modules: { router },
  view: (state, actions) =>
    <main>
      <Route path="/" exact="true" view={Home(state, actions)} />
      <Route path="/adventure" view={Away} />
    </main>
  ,
})
```

I have not got an example of data fetching.. although I am starting to thing that this might not be the routers responsibility.. rather that of a dedicated `Fetch` component. that can live inside of route views. That will probably be the next component I try build.

src/components/Popstate.js
```diff
@@ -0,0 +1,20 @@
+let setup = false
+export const Popstate = ({ update }) => {
+  if (!setup) {
+    setup = true
+    // Update path when a pushState occurs
+    const pushState = history.pushState
+    history.pushState = function() {
```
SkaterDad: Can these `history` modifications create conflicts with other libraries or unintended bugs?

This is a pretty clever trick to get around passing the `go` function around.

JorgeBucaran:
@lukejacksonn Can you fix the git conflict?

LICENSE.md
```diff
@@ -1,19 +0,0 @@
-Copyright © 2017-present [Jorge Bucaran](https://github.com/jbucaran)
```
JorgeBucaran: @lukejacksonn Fix the LICENSE as we discussed.

.travis.yml
```diff
@@ -1,17 +0,0 @@
-language: node_js
```
JorgeBucaran: @lukejacksonn Please don't remove this file. 😕

.editorconfig
```diff
@@ -1,9 +0,0 @@
-root = true
```
JorgeBucaran: Nice! 👍

src/utils.js
```diff
@@ -0,0 +1,38 @@
+export const extractParams = route => {
```
JorgeBucaran: @lukejacksonn Query is nice, but I recommend we work on this after the new  router is published, so that it's done right and better researched.

rollup.config.js
```diff
@@ -1,12 +1,23 @@
-import uglify from "rollup-plugin-uglify"
+import babel from 'rollup-plugin-babel'
+import uglify from 'rollup-plugin-uglify'
 
 export default {
```
JorgeBucaran: @lukejacksonn Can you completely remove this file? And use the same approach on Hyperapp. See Hyperapp's  package.json.

src/components/Popstate.js
lukejacksonn: There has been discussion about this on slack too (you really should join us!) and as you can see, it leaves `pushState` and `popState` exactly as is, but extends it (ensures an action is called subsequently). The only real risk I can see here is if the action for some reason ceases to exist.. to guard agains that `typeof update === 'function' && update()` was proposed.

src/utils.js
lukejacksonn: What is the advantage of doing this later and what is your concern with this implementation?

I'd argue any querystring parsing is better than none. I can clearly document which cases this will work for and for which it will fail. If anyone wants to propose an improvement it then they can once this gets merged.

JorgeBucaran: @lukejacksonn You are moving too fast. When you move fast things get out of control quickly. Users will come in asking for new features before we are even sure we even need this or not. Then we might break the API. If we push only the bare minimum we'll be safe.

JorgeBucaran: @lukejacksonn I am not sure we need a qs in the router.

SkaterDad: In my opinion the router needs to handle query string parsing.  At the very least, allow users to pass in their own function as a configuration option.

JorgeBucaran:
What's up with this? GitHub? 😆 

> This branch has conflicts that must be resolved.

src/utils.js
SkaterDad: @JorgeBucaran Was there a specific reason you didn't feel a qs parser was needed in the router?

LICENSE.md
```diff
@@ -1,19 +1,8 @@
...
+Copyright © 2017 [Jorge Bucaran](https://github.com/JorgeBucaran)
+Copyright © 2017-present [Luke Jackson](https://github.com/lukejacksonn)
```
JorgeBucaran: @lukejacksonn It's missing a new line. 😆

src/utils.js
JorgeBucaran: Not particularly, I was just trying to simplify the PR so we could merge it more quickly. I'll have a look at this file and leave comments then. 👍

src/components/Route.js
```diff
@@ -0,0 +1,14 @@
+import { extractParams, extractQuery } from '../utils'
+
+export const Route = ({ exact, path = '', view }) => {
+  const match = exact
```
JorgeBucaran: @lukejacksonn The Route component tries to match the pathname with the given path (and if there is match extractParams and extractQuery) **every time**.

How do you feel about that? Seems wasteful. I think a `Router` parent component could help ameliorate this. What is the advantage of only `Route`?

lukejacksonn: Yes, match gets calculated every render, not sure how expensive it is but I assume the overhead is less than or the same as that of calling `h` which doesn't really concern me. The more expensive function `extractParams` and `extractQuery` only get called when required (match is true). If perf was a concern here you could memoize any of the 3 functions, but I don't see a lot of gain to be had.

The benefit you get from dynamic routing like this, rather than static routing like this old router is flexibility and composability. Routing is a hard problem and there are often complex requirements; not always a simple 1-to-1 mappings of routes to views. If you can meet these requirements with various configurations of very simple components like `Route`, `Switch` and `Link` then why introduce another thing?

A parent `Router` component implies you are to be confined within that component, it basically takes over your app, becoming the top component. It demands nesting of children but could it be nested itself like a simple route? I have not thought about it too much because I haven't come across a need for one. How does it help?

Issue #27, opened 2017-11-17
A different API.


JorgeBucaran:
@lukejacksonn's #26 and @speier's #23 are the best we have at the moment, but I am still not satisfied with the current approach, so here is (yet) another take on the router's API.

I don't have a PR ready yet, but I wanted to share my ideas right away. Here is an example of how your code would look like.

```js
import router, { Router } from "@hyperapp/router"
```

### Style I

```js
app({
  state: {
    router: router.state
  },
  actions: {
    router: router.actions
  },
  view: Router([["/", indexView], ["/", adventureView]])
})
```
  

### Style II

```js
app({
  state: {
    router: router.state
  },
  actions: {
    router: router.actions
  },
  view: Router([
    {
      path: "/",
      view: indexView
    },
    {
      path: "/",
      view: adventureView
    }
  ])
})
```

SkaterDad:
Your proposed API seems good for simple cases, much like the original built-in router.  It would be great to get something released that actually works so new people aren't put off.

Do you want this package to support more complex use cases, which need guarding/redirects/data/etc...?  Or just keep it simple?

I've got a much more complicated router in the works for my own use cases, which I may end up publishing once my use case is done dictating the API & features..

JorgeBucaran:
@SkaterDad I definitely want to support all of that stuff, let me think how that would look like with this API. 

What would those features be and can you define them?

JorgeBucaran:
@SkaterDad Isn't redirects === guarding? 🤔

Swizz:
`Router` will return a vdom tree, isnt it ? So I will be able to use it into a template ?

```js
app({
  state: {
    router: router.state
  },
  actions: {
    router: router.actions
  },
  view: (state, actions) => main([
    Router([["/", indexView], ["/", adventureView]])(state, actions)
  ])
})
```

JorgeBucaran:
@SkaterDad What is a template?

Swizz:
A wrong name for "used into a vdom tree, as a child"

JorgeBucaran:
@Swizz It returns the view we must give to app() so your app is rendered. So, you don't need to use it like that; it's done for you.

Swizz:
Okay then. I have an header (a nav bar) part of the app. But I only want the Router to deal with the section content. I need to use the header as a stateless component on each view ?
I think that is why both @lukejacksonn and @speier are awaiting Router component.

[...]

PR #28, opened 2017-12-01
New router API.

JorgeBucaran:
The new API is very similar to ReactRouter, so if you are familiar with ReactRouter you already know how most of this works.

## `location` and setup

The router exposes a `location` module. Wire it to your app's state and actions so we can map a subset of [window.location](https://developer.mozilla.org/en-US/docs/Web/API/Location).

```jsx
const actions = app({
  state: {
    location: location.state
  },
  view,
  actions: {
    location: location.actions
  }
})

const unsubscribe = location.subscribe(actions.location)
```

### location.subscribe

Subscribe to popstate and pushstate (this one doesn't really exist, so we need to create it behind the scenes).

```js
const unsubscribe = location.subscribe(actions.location)
```

### unsubscribe

Call `unsubscribe` to remove event listeners and undo our custom pushstate event.

```js
unsubscribe()
```

## Route

```jsx
<Route path="/my-path" render={({ match, location }) => <h1>Hello</h1>} />
```

#### location
- The location state. 
##### location.pathname
- Same as window.location.pathname.
##### location.previous
- The previous location.pathname. Useful when redirecting back to the referrer url/pathname after leaving a guarded/protected route.

#### match.url
- The matched part of the url. Use to assemble links inside routes. See [Link](#link).
#### match.path
- The route path. Same as `path`.
#### match.isExact
- Indicates whether the given path matched the url exactly or not. I am still debating whether we need this or not.

## Link

```jsx
<Link to="/hello">Hello Page</Link>
<Link to="/public">Public Page</Link>
<Link to="/private">Private Page</Link>
```

## Switch
```jsx
<Switch>
  <Route path="/" render={Home} />
  <Route path="/will-match" render={WillMatch} />
  <Route render={NoMatch} />
</Switch>
```

## Redirect
```jsx
<Route
  path="/protected"
  view={() => (isLoggedIn ? <Protected /> : <Redirect to="/login" />)}
/>
```

### Basic
```jsx
const homeView = () =>
  <div>
    <h2>Home</h2>
  </div>

const aboutView = () =>
  <div>
    <h2>About</h2>
  </div>

const topicView = ({ match }) => {
  return (
    <div>
      <h3>
        You are reading: {match.params.topicId}
      </h3>
      <hr />
      <h3>Subtopics</h3>
      <ul>
        <li>
          <Link to={`${match.url}/red`}>Red</Link>
        </li>
        <li>
          <Link to={`${match.url}/blue`}>Blue</Link>
        </li>
      </ul>

      <Route path={`${match.path}/red`} render={() => <h1>RED</h1>} />
      <Route path={`${match.path}/blue`} render={() => <h1>BLUE</h1>} />
    </div>
  )
}

const topicsView = ({ match }) => {
  return (
    <div>
      <h2>Topics</h2>
      <ul>
        <li>
          <Link to={`${match.url}/one`}>Hyperapp Topic One</Link>
        </li>
        <li>
          <Link to={`${match.url}/two`}>Hyperapp Topic Two</Link>
        </li>
        <li>
          <Link to={`${match.url}/three`}>Hyperapp Topic Three</Link>
        </li>
      </ul>

      {match.isExact && <h3>Please select a topic.</h3>}

      <Route parent path={`${match.path}/:topicId`} render={topicView} />
    </div>
  )
}

const view = state =>
  <div>
    <ul>
      <li>
        <Link to="/">Home</Link>
      </li>
      <li>
        <Link to="/about">About</Link>
      </li>
      <li>
        <Link to="/topics">Topics</Link>
      </li>
    </ul>

    <hr />

    <Route path="/" render={homeView} />
    <Route path="/about" render={aboutView} />
    <Route parent path="/topics" render={topicsView} />
  </div>
```

### Url Params
```jsx

const view = state =>
  <main>
    <h2>Accounts</h2>
    <ul>
      <li>
        <Link to="/netflix">Netflix</Link>
      </li>
      <li>
        <Link to="/zillow-group">Zillow Group</Link>
      </li>
      <li>
        <Link to="/yahoo">Yahoo</Link>
      </li>
      <li>
        <Link to="/modus-create">Modus Create</Link>
      </li>
    </ul>

    <Route
      path="/:id"
      render={({ match }) =>
        <div>
          <h3>
            ID: {match.params.id}
          </h3>
        </div>}
    />
  </main>
```

### Redirects (Auth)
```jsx

const fakeAuth = {
  isAuthenticated: false,
  authenticate(cb) {
    fakeAuth.isAuthenticated = true
    setTimeout(cb, 2000) // fake async
  },
  signout(cb) {
    fakeAuth.isAuthenticated = false
    setTimeout(cb, 2000)
  }
}

const AuthButton = ({ go, isLoggedIn, userId, login }) =>
  isLoggedIn
    ? <p>
        Welcome {userId}!
        <button
          onclick={() => {
            fakeAuth.signout(() => {
              go("/")
              login(null)
            })
          }}
        >
          Sign out
        </button>
      </p>
    : <p>You are not logged in.</p>

const Public = () => <h3>Public</h3>

const Protected = () => <h3>Protected</h3>

const Login = ({ from, login, redirectToReferrer }) => props => {
  if (redirectToReferrer) {
    return <Redirect to={from} />
  }

  return (
    <div>
      <p>
        You must log in to view the page at {from}
      </p>
      <button
        onclick={() => {
          fakeAuth.authenticate(() => login("Admin"))
        }}
      >
        Log in
      </button>
    </div>
  )
}

const AuthExample = ({
  location: { go, previous },
  login,
  userId,
  isLoggedIn,
  data
}) => {
  return (
    <main>
      <AuthButton
        go={go}
        login={login}
        userId={userId}
        isLoggedIn={isLoggedIn}
      />

      <ul>
        <li>
          <Link to="/public">Public Page</Link>
        </li>
        <li>
          <Link to="/protected">Protected Page</Link>
        </li>
      </ul>

      <Route path="/public" render={Public} />

      <Route
        path="/protected"
        render={() => (isLoggedIn ? <Protected /> : <Redirect to="/login" />)}
      />

      <Route
        path="/login"
        render={Login({
          login,
          redirectToReferrer: isLoggedIn,
          from: previous
        })}
      />
    </main>
  )
}

const view = state => actions =>
  <AuthExample
    location={{
      go: actions.location.go,
      previous: state.location.previous
    }}
    login={actions.auth.login}
    userId={state.auth.userId}
    isLoggedIn={state.auth.userId !== null}
  />
```

### Recursive
```jsx
const PEEPS = [
  { id: 0, name: "Michelle", friends: [1, 2, 3] },
  { id: 1, name: "Sean", friends: [0, 3] },
  { id: 2, name: "Kim", friends: [0, 1, 3] },
  { id: 3, name: "David", friends: [1, 2] }
]

const find = id => PEEPS.find(p => p.id == id)

const view = state =>
  <Person
    match={{
      url: "",
      params: { id: 0 }
    }}
  />

const Person = ({ match }) => {
  const person = find(match.params.id)

  return (
    <div>
      <h3>
        {person.name}’s Friends
      </h3>
      <ul>
        {person.friends.map(id =>
          <li key={id}>
            <Link to={`${match.url}/${id}`}>
              {find(id).name}
            </Link>
          </li>
        )}
      </ul>
      <Route parent path={`${match.url}/:id`} render={Person} />
    </div>
  )
}
```

### No Match (404)

```jsx
const NoMatchExample = () =>
  <div>
    <ul>
      <li>
        <Link to="/">Home</Link>
      </li>
      <li>
        <Link to="/old-match">Old Match, to be redirected</Link>
      </li>
      <li>
        <Link to="/will-match">Will Match</Link>
      </li>
      <li>
        <Link to="/will-not-match">Will Not Match</Link>
      </li>
      <li>
        <Link to="/also/will/not/match">Also Will Not Match</Link>
      </li>
    </ul>
    <Switch>
      <Route path="/" render={Home} />
      <Route
        path="/old-match"
        render={() => <Redirect from="/old-match" to="/will-match" />}
      />
      <Route path="/will-match" render={WillMatch} />
      <Route render={NoMatch} />
    </Switch>
  </div>

const view = state => <NoMatchExample />

const Home = () =>
  <p>
    A <code>&lt;Switch></code> renders the first child <code>&lt;Route></code>{" "}
    that matches. A <code>&lt;Route></code> with no <code>path</code> always
    matches.
  </p>

const WillMatch = () => <h3>Matched!</h3>

const NoMatch = ({ location }) =>
  <div>
    <h3>
      No match for <code>{location.pathname}</code>
    </h3>
  </div>
```


okwolf:
@JorgeBucaran I thought we were using [`Navigation`](https://github.com/hyperapp/router/issues/27#issuecomment-345746925) instead of `Switch` for that component.

JorgeBucaran:
@okwolf The Switch component mentioned on that thread refers to a different component, and the one in this PR is more like the Switch component in ReactRouter.

okwolf:
> @JorgeBucaran The Switch component mentioned on that thread refers to a different component, and the one in this PR is more like the Switch component in ReactRouter.

I just looked at the docs for `<Switch>` from `react-router` and now I see where the name came from. I still think the name is poorly-chosen and ambiguous, but if it's familiar to users of `react-router` then I'll keep my virtual mouth shut on the subject 😐

JorgeBucaran:
@okwolf Do you have a better name than Switch?

okwolf:
> @JorgeBucaran Do you have a better name than Switch?

I really wish I did, but I don't.

Building elegant software is so easy compared to naming things... 😒

yuku-t:
@JorgeBucaran It seems that `<Link>` ignores all properties except `to` and `location` so there is no way to set attributes (such as `class`) to underlying `<a>` node. I think it isn't intended behavior.

JorgeBucaran:
@yuku-t Good catch, thanks! 👍

JorgeBucaran:
Now you can set any attribute of the `<a>`, but maybe it would be better to decide what attributes of the hyperlink we would want to set, e.g.,`id`, `class`, etc.

1zaak:
Hey @JorgeBucaran, first of all, I am coming from the Angular world, and just starting to understand React/Redux, etc.  I am just playing around with Hyperapp and I must say, this is gonna be (if not is) the next big thing, what an awesome project and community. Thanks!

My problem is, I'm trying to use the new router here (with Hyperapp's v0.16.0) by pointing to this branch. And I'm having trouble "reloading" the app. It only reloads (and renders the view) after _another_ action was triggered. Could you explain a bit more on how to configure the new location module of the router at the start of this thread? I think I may have missed something there.

Looking forward to understanding all of Hyperapp!

JorgeBucaran:
@1zaak Thanks for the kind words. Here is a simple (but full code) example.

```jsx
import { h, app } from "hyperapp"
import { Link, Route, location } from "@hyperapp/router"

const actions = app({
  state: {
    auth: { userId: null },
    location: location.state
  },

  view: state =>
    <main>
      <h2>Accounts</h2>
      <ul>
        <li>
          <Link to="/netflix">Netflix</Link>
        </li>
        <li>
          <Link to="/zillow-group">Zillow Group</Link>
        </li>
        <li>
          <Link to="/yahoo">Yahoo</Link>
        </li>
        <li>
          <Link to="/modus-create">Modus Create</Link>
        </li>
      </ul>

      <Route
        path="/:id"
        view={({ match }) =>
          <div oncreate={() => console.log("OK")}>
            <h3>
              ID: {match.params.id}
            </h3>
          </div>}
      />
    </main>,

  actions: {
    auth: {
      login: userId => ({ userId })
    },
    location: location.actions
  }
})

const unsubscribe = location.subscribe(actions.location)
```

Of course, the router is still not published, but if you managed to import the router from this branch, then the above should run without any trouble.

1zaak:
@JorgeBucaran It's still not working. Even after I copy the whole thing.

JorgeBucaran:
@1zaak Well surely it's a problem with how you are importing from this branch or with the hyperapp build you are using because as you can see from the Gif, it is actually working. (hyperapp v0.16.2)

You'll need to give me a few more details if you want me to help (and I'll try to do what I can).

1zaak:
Ah, I was using v0.16.0. It works after i upgraded hyperapp tov0.16.2. Thanks a lot @JorgeBucaran

SkaterDad:
@JorgeBucaran What is your recommended way to fetch data for routes in this new router API?

JorgeBucaran:
@SkaterDad What about using oncreate on the component that will be rendered?

```jsx
<Route route="/" view={Hello} />
```
```jsx
const Hello = props => (
  <main key="hello" oncreate={() => { fetch }}
  </main>
)
```

What else did you have in mind?

src/Switch.js
```diff
@@ -0,0 +1,3 @@
+export function Switch(props, children) {
+  return children[0]
```
lukejacksonn: How does this work 🤔

SkaterDad:
> @JorgeBucaran What about using oncreate on the component that will be rendered?

I suppose that would work for many use cases, so long as `key` is used thoughtfully.  It is a clean solution to preventing extra fetches on re-render, and you could change the key if your data needs re-fetching.

In that case, changing the key to fetch fresh data, would all of the children nodes get wiped out and rendered from scratch?  Could be a perf hit when all you really need is to update data of the same shape.

src/Switch.js
vdsabev: I was wondering the same thing. Shouldn't it be the first **matched** child?

vdsabev: Oh, wait, `<Route />` only renders if the path matches, so I think the `children` array would really have only 1 item.

JorgeBucaran:
@SkaterDad 

> In that case, changing the key to fetch fresh data, would all of the children nodes get wiped out and rendered from scratch? 

If the data has the same shape and so has the UI tree, then nodes will not be wiped out and rendered from scratch. Otherwise, you can key the children, so that they are only moved around.

Does that make sense?

src/Switch.js
JorgeBucaran: @lukejacksonn @vdsabev Yes, exactly. Switch ensures at least one element is rendered and if no route matches, then `children[0]` will be undefined, so it works.

src/Route.js
```diff
@@ -0,0 +1,19 @@
+import { parseRoute } from "./parseRoute"
+
+export function Route(props) {
+  var location = props.location || window.location
+  var path = props.path
+  var match =
+    !path ||
+    parseRoute(path, location.pathname, {
+      exact: !props.parent
```
leavehouse: Is the idea here that anyone who wants non-exact route matching will use `parent` even if the component being routed to has no child routes?

JorgeBucaran: @leavehouse Yes. Whether the component declares child routes or not is left to the user.

1zaak:
I encountered an issue when using Switch. The way I implement before I added routes was like so:
```
export const UsersRow = ({row}) => {
    return <div class="columns">
        {row.map(({id, name, age})=>{
        return (<div>
            <UserCard id={id} name={name} age={age}/>
            </div>)
        })}
    </div>
}

export const Users = ({user}) => {    
    let rows = chunk(users, 4)
    return rows.map(row=>{
        return <UsersRow row={row}/>
    })
}
```

And rendering in view.

```
<Products products={state.products}/>
```

But when I added routes.
```
<Switch>
    <Route path="/users" render={Products({products: state.products})} />
      <Route>...redacted...
</Switch>
```

And changed the signature of the Users view component to add props:
```
export const Users = ({user}) => props => {    
    let rows = chunk(users, 4)
    return rows.map(row=>{
        return <UsersRow row={row}/>
    })
}
```
 
It renders correctly except for the UsersRow which fails to render. Am i missing something here?

JorgeBucaran:
@1zaak Switch renders the first child, that's it! Switch is unique in that it renders a route exclusively. So, maybe you were using Switch, but didn't have to?

1zaak:
That's what I thought, but I think I tested it without Switch and still the same thing happened. But let me try again, maybe I did something wrong. Thanks for confirming the right behavior @JorgeBucaran

JorgeBucaran:
@1zaak Let me know how that goes! Remember that every Route that matches the location renders inclusively, but if inside a Switch, only the first will be matched. It should work similarly as ReactRouter's Switch component. Also make sure you didn't miss the 404 example using Switch above on my post (it's the last of the examples). 💪

1zaak:
This works for me (without Switch)

```
<Route path="/users" render={()=>{
            return isAuthenticated() && <Users products={state.users}/>
}} />
```

Thanks @JorgeBucaran

PR #32, opened 2017-12-21
Allow using Link without Hyperapp

vdsabev:
## Why is this useful?
I'm using the router with picodom instead of Hyperapp with great success - it does outstanding work for such a small router library, and plays well with other architectures. The only component that doesn't work out of the box is `Link`, because it imports `h` from Hyperapp.

## What's the workaround?
I've basically copied all of the code from `Link.js` into my own application, with the only difference being I'm importing a different `h` function - the one from picodom.

## Is there a better way?
If `Link` could somehow receive `h` as a parameter, it would completely decouple it from Hyperapp and allow everyone to use it however they want.

## What did I do?
I moved most of the code from `Link.js` to a new script called `createLink.js`. It exports a function with a single parameter - `h`, which returns what used to be in `Link.js`. So we can now do this in `Link.js` instead:
```js
import { h } from "hyperapp"
import { createLink } from "./createLink"

export var Link = createLink(h)
```

At the same time, in my own application, I can use picodom with the same effort:
```js
import { h } from 'picodom';
import { createLink } from '@hyperapp/router/src/createLink';

export const Link = createLink(h);
```

## What about other libraries?
There's really nothing limiting this to picodom. Any other VDOM library should work great, and I don't see why you can't use that without VDOM, probably even with vanilla JS, by passing a custom render function:
```js
export const Link = createLink((tag, props, children) => {
  const el = document.createElement(tag);
  for (let key in props) {
    el[key] = props[key];
  }
  // append the children somehow...out of scope for this PR 😋

  return el;
});
```


vdsabev:
@jorgebucaran can we reconsider this PR? The solution I described is still valid as of 0.5.1. It would still be useful to reduce the dependency on hyperapp and avoid copy/pasting code to use `Link` without relying on `h`, as you have in 48d81a0cca9ce0545905021b1cd78d9d4840a33d.

Let me know what you think and I can redo the PR later today 🤠

jorgebucaran:
@vdsabev If it won't be much trouble, please make a new PR so we can review it again! :)

vdsabev:
Looks like both #36 and #61 are open and contain changes to `Link.js`. In the interest of avoiding conflicts, should I wait for those to get merged or closed?

jorgebucaran:
@vdsabev You could wait, or send your PR and update/resolve conflicts as those other two get resolved. What do you think? It's up to you.

PR #35, opened 2018-01-03
Add hash router capability

jrop:
This is an updated PR in the same vein as #1 and has the following features:

- backwards compatible with the current API
- `Link`/`Route` are created using a factory that takes as input a "provider".  These providers provide methods to access/mutate/subscribe to different location backends (in this case `window.location.pathname`/`pushState` and `window.location.hash`)
- The default exported `Link`/`Route` are tied to the `window.location` backend

To use with hash routing (for example):

## index.html (place at project root to test)
```html
<div id="root"></div>
<script src="https://unpkg.com/hyperapp@1.0.1/dist/hyperapp.js"></script>
<script src="https://unpkg.com/@jrop/hyperx@2.3.2/dist/hyperx.min.js"></script>
<script src="dist/router.js"></script>
<script>
	const { app, h } = hyperapp
	const hx = hyperx(h)

        // obtain specialized versions of Link/Route:
	const { Link, Route } = router.hash
	const STATE = Object.assign({}, router.hash.state)
	const ACTIONS = Object.assign({}, router.hash.actions)
	const main = app(STATE, ACTIONS, () => hx`<div>
		<h1>App</h1>
		${Link({ to: '/' }, 'Home')}
		${Link({ to: '/about' }, 'About')}

		${Route({ path: '/', render: () => hx`<p>Home</p>` })}
		${Route({ path: '/about', render: () => hx`<p>About</p>` })}
	</div>`, document.getElementById('root'))
	const unsubscribe = router.hash.subscribe(main)
</script>
```

The two included "providers" are properties on the exported `router` and have the following shape:

```js
router = {
    location: {
        state: ..., actions: ...,
        subscribe: ...,
        Link: ..., Route: ...
    },
    hash: {
        state: ..., actions: ...,
        subscribe: ...,
        Link: ..., Route: ...
    },
    ...
}
```

Before and after file sizes:
**Before**: router.js: 941 bytes (gzip'd)
**After**: router.js: 1159 bytes (gzip'd)


JorgeBucaran:
Sorry, @jrop, I had an issue with my GPG credentials and had no choice, but to push --force to master and it messed up the PR. 🙇 

Could you please create a new PR for your changes?

jrop:
Will do

Issue #30, opened 2017-12-20
Switch requires JSX to work


zaceno:
See this example:

https://codepen.io/zaceno/pen/OzXxjb?editors=0010

It renders `<h1>Bar</h1>` correctly according to the routing rules.  It uses a view function (`viewJSX`) containing a `<Switch>...</Switch>` in JSX, so it works as it should.

Now comment out `viewJSX` in the app-statement, and un-comment `viewFunc`.  It is the exact same function, only it calls the switch as a function, using `Switch({}, [...])`.

Notice that it doesn't work now. It renders nothing. This is because it is rendering the first result of it's children, which is `false` (because the first child's `path` is not a match for the current location)

JSX filters out falsy children automatically, which is why it works with JSX.

JorgeBucaran:
@zaceno You are on fire! 🔥 

Jut to clarify: wasn't it `Switch({}, ...)` incorrect and `h(Switch)` the preferred way. I can't remember.

I think this is probably a bug anyway, so I'll tag it as such and look into it later.

Thanks for the report! 🙏

zaceno:
>  wasn't it Switch({}, ...) incorrect and h(Switch) the preferred way. I can't remember.

Well, I've never used a function like `h(Component, props, children)`, and haven't seen anyone else do it either that I can remember. It's always `Component(props, children)`. And I much prefer that convention. 

But good point: `h(Switch, {}, child1, child2, child3)` is more accurately how JSX is calling it. And I think 

> JSX filters out falsy children automatically, which is why it works with JSX.

 is probably incorrect. It's `h` that does the filtering.

zaceno:
Fixed!

PR #39, opened 2018-01-15
Allow Switch component without h

zaceno:
Switch should return only the first match of its children, i e only the first of non-falsy children.

Previous implemetation relied on calling as: `h(Switch, null, [....])` because h filters out falsy children. 

This implementation does not assume all children are truthy, allowing you to call it simply as: `Switch(null, [...])`

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -1,3 +1,5 @@
 export function Switch(props, children) {
-  return children[0]
+  var i = 0
+  while (!children[i] && i < children.length) i++
+  return children[i]
 }
--- a/test/switch.test.js
+++ b/test/switch.test.js
@@ -1,13 +1,13 @@
 import { Switch } from "../src/Switch"
 
-test('Switch returns only the first truthy child', () => {
-  const expected = 'truthy value'
-  const children = [0, null, expected, false, 'other']
+test("Switch returns only the first truthy child", () => {
+  const expected = "truthy value"
+  const children = [0, null, expected, false, "other"]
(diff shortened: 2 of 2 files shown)


zaceno:
Fixes:  https://github.com/hyperapp/router/issues/30

JorgeBucaran:
@zaceno Oh, that makes sense. Thank you for finally bringing sense to this issue. 👏

JorgeBucaran:
@zaceno The code is equivalent to this correct?

```js
for (var i = 0; i < children.length; i++) {
  if (children[i]) {
    return children[i]
  }
}
```

zaceno:
Yes should be equivalent!

JorgeBucaran:
@zaceno Could add a test please?

zaceno:
Hehe, had a feeling you'd ask... yeah alright. 😄  Might be a day or two though.

zaceno:
Sorry... I forgot you like strings with double quotes 😛  Fixed that.

okwolf:
@zaceno doesn't this project also have a format script? 😕

zaceno:
@okwolf ah, hehe, good call. Yes it does. I just forgot all about it. But I ran it now and it produced no diff -- so fixing the quotes apparently was my only stylistic mishap.

JorgeBucaran:
> @zaceno  so fixing the quotes apparently was my only stylistic mishap.

Haha. Thanks for adding tests! 👍

JorgeBucaran:
🎉 🎉

Issue #38, opened 2018-01-14
`oncreate` doesn't get called when you route to a new View


ForsakenHarmony:
I have 2 views, `/` (home) and `/search`.

in search I have `oncreate={actions.search.findNearby}`, this doesn't get called when I route to it from `/`, but it does when I reload the page

zaceno:
It's probably because the vdom engine is repurposing the top elements from the two views. Try setting a different key on the two views top nodes. 

Actually... that will only work if the route components themselves are enclosed in a vnode because keys have no effect on the topmost vnode of an app.

ForsakenHarmony:
uhh, ok?

zaceno:
If my response didn't make sense to you, perhaps you could post a simplified version of the code you have that isn't working, and we can just show you how to make it work.

The problem you're describing is not uncommon and I'm pretty sure it can be solved by keying your different views.

ForsakenHarmony:
Yeah, it worked, I was confused about the second part of what you said, makes sense now though.

Maybe the router could add the path as a key by default

ForsakenHarmony:
misclick, I can make a pr for that if you want it, or you can just close this

zaceno:
> Maybe the router could add the path as a key by default

That's a pretty neat idea! It's the key you'd naturally choose anyway. 

PR away by all means (but it's not my project and I can't speak for anyone else ;) )

ForsakenHarmony:
You're a member ¯\_(ツ)_/¯

JorgeBucaran:
> Maybe the router could add the path as a key by default

How?

@zaceno Could you explain what they are saying? 🙏

ForsakenHarmony:
Just add a `key` to the `Route` with the `path` for the route

ForsakenHarmony:
I found a way

```js
import { parseRoute } from "./parseRoute"

export function Route(props) {
  var location = props.location || window.location
  var match = parseRoute(props.path, location.pathname, {
    exact: !props.parent
  })

  if (!match) return;

  const child = props.render({
		match: match,
		location: location
	});

  child.props.key = child.props.key || props.path;

  return child;
}
```

JorgeBucaran:
@ForsakenHarmony If this works, then I am 💯 in! 

Could you create a PR with tests? 🙏

ForsakenHarmony:
unsure how to test it

JorgeBucaran:
Well, I can think about it then. Feel free to make that PR anyway! 👍

PR #40, opened 2018-01-15
Add key to the element wrapped in the Route

ForsakenHarmony:


package.json
```diff
@@ -24,12 +24,15 @@
     "bundle": "rollup -i src/index.js -o dist/router.js -f umd -mn router -g hyperapp:hyperapp",
     "minify": "uglifyjs dist/router.js -o dist/router.js -mc pure_funcs=Object.defineProperty --source-map includeSources,url=router.js.map",
     "prepublish": "npm run build",
-    "format": "prettier --semi false --write 'src/**/*.js'",
+    "format": "prettier --semi false --write \"src/**/*.js\"",
```
JorgeBucaran: @ForsakenHarmony Are you on windows? What's the purpose of this? Just curious.

src/Route.js
```diff
@@ -6,11 +6,14 @@ export function Route(props) {
...
-    })
-  )
+  if (!match) return
+
+  var child = props.render({
+    match: match,
+    location: location
+  })
+
+  child.props.key = child.props.key || props.path
```
JorgeBucaran: @ForsakenHarmony @zaceno Is this infallible? 

Anyone familiar with ReactRouter, what's going on there? 

I like this though.

package.json
```diff
@@ -24,12 +24,15 @@
...
     "release": "npm run build && npm test && git commit -am $npm_package_version && git tag $npm_package_version && git push && git push --tags && npm publish"
   },
   "babel": {
     "presets": "env"
   },
+  "jest": {
+    "testURL": "http://test/test"
```
JorgeBucaran: Just leaving a comment here: https://facebook.github.io/jest/docs/en/configuration.html#testurl-string

Nice.

src/Route.js
ForsakenHarmony: `props` always gets added by `h` at least, if that's what you're asking

package.json
ForsakenHarmony: I wanted to do it in the test, but couldn't figure out how to do that

package.json
```diff
@@ -24,12 +24,15 @@
     "bundle": "rollup -i src/index.js -o dist/router.js -f umd -mn router -g hyperapp:hyperapp",
     "minify": "uglifyjs dist/router.js -o dist/router.js -mc pure_funcs=Object.defineProperty --source-map includeSources,url=router.js.map",
     "prepublish": "npm run build",
-    "format": "prettier --semi false --write 'src/**/*.js'",
+    "format": "prettier --semi false --write \"src/**/*.js\" \"test/**/*.js\"",
```
ForsakenHarmony: I used normal quotes, because single quotes seemed to break it on windows

src/Route.js
JorgeBucaran: Nope, I mean does ReactRouter also add a key to the route component?

ForsakenHarmony: afaik, no, but they have component methods, which don't depend on elements that might be reused

lukejacksonn:
Could you provide a description explaining what this PR does and what problem you are trying to solve?

ForsakenHarmony:
#38

lukejacksonn:
Thanks! So just to clarify the desired behaviour; in this case:
```
main([
  Route({ render: Nav }),
  Route({ path: '/users', render: () => div({ key: 'users' }, 'Some Users:')  }),
  Route({ path: '/users/:id', render: User }),
])
```

The first route gets no key
The second inherits its key `users`
The third route gets the key of `/users/:id`

Is this correct?

ForsakenHarmony:
Yea, I guess so

src/Route.js
zaceno: @JorgeBucaran is it infallible? ...well... this:

```jsx
<main>
   <Route path="/show/secrets" render={SecretComponent} />
   <AComponent />
   <AnotherComponent />
   <Route path="/show/secrets" render={AnotherSecretComponent} />
   <ThirdComponent />
</main>
```

I e using the route as a switch to turn on or off multiple sibling components. Then the siblings will be keyed with the same key which of course is no good. ...unless the components have their own keys, in which case those keys will be used instead. So:

A - Before: By default `oncreate` won't get called for routed components. You have to manually key them to make it work. Beginners are required to understand the relationship between keys and lifecycle events. Higher threshold.

B - This proposal: You get automatic keys which make sense for the most part. And hence a lower threshold for beginners. But you run the risk of key-collisions and you don't understand why.

lukejacksonn:
The path seems like a sensible key to use. It is a good identifier for the element. The only doubt I have; is it guaranteed to be unique? Does it even matter if it is not? If it does matter, how should we document it? Otherwise SGTM 👍

src/Route.js
JorgeBucaran: What about this?

```js
var key= props.key || (props.path + (props.render ? props.render.name + props.render.length : ""))
```

ForsakenHarmony:
it was mostly to address an issue I had with element reuse and `oncreate` not being called

would be the same thing with non unique paths ig

JorgeBucaran:
@lukejacksonn @ForsakenHarmony Zaceno explained it best [here](https://github.com/hyperapp/router/pull/40#discussion_r161605880).

src/Route.js
JorgeBucaran: Arrow functions have no name. 😞

ForsakenHarmony: They do if you assign them to a variable, but anonymously used, they don't, just `""`

JorgeBucaran: I guess we could use `props.render.toString()`, but that feels asking for trouble.

PR #46, opened 2018-02-11
add typescript type definitions

m0a:
add typescript type definitions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,40 @@
+import { VNode } from "hyperapp";
+
+/** Link */
+interface LinkProps {
+  to: string;
+  location?: Location;
+}
+export function Link(props: LinkProps): VNode<LinkProps>;
+
+/** Route */
+interface RouteProps {
+  parent: any;
+  path: string;
+  render: () => VNode<any>;
+}
+export function Route(props: RouteProps): any;
+
+/**Switch */
+export function Switch(props: any, children: VNode<any>): VNode<any>;
+
+/** Redirect */
+type RedirectProps = LinkProps;
+export function Redirect(props: RedirectProps): VNode<RedirectProps>;
+
+/** location */
+interface LocationState {
+  pathname: string;
+  previous: string;
+}
+interface LocationActions {
+  go: (pathname: string) => any;
+  set: (data: any) => any;
+}
+interface Location {
+  state: LocationState;
+  actions: LocationActions;
+  subscribe: (actions: LocationActions) => void;
+}
+
+export declare const location: Location;
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,7 +3,7 @@ import { VNode } from "hyperapp";
 /** Link */
 interface LinkProps {
   to: string;
-  location?: Location;
+  location?: LocationState;
 }
 export function Link(props: LinkProps): VNode<LinkProps>;
 
@@ -15,7 +15,7 @@ interface Match<P> {
   params: P;
 }
 interface RenderProps<P> {
-  location: Location;
+  location: LocationState;
   match: Match<P>;
 }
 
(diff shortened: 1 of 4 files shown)


m0a:
sorry. this type definition is wrong
I will re-submit the re-corrected one

JorgeBucaran:
@m0a Thanks.

petersondrew:
Is the package.json conflict the only thing holding back this PR @m0a?

bpryer:
Any reason this PR has not been merged?

m0a:
Thank you for review @joseluisq 
but RouteProps.parent was already optional.
https://github.com/m0a/router/blob/master/index.d.ts#L23

index.d.ts
```diff
@@ -8,19 +8,30 @@ interface LinkProps {
...
+  params: P;
+}
+interface RenderProps<P> {
+  location: Location;
+  match: Match<P>;
+}
+
 interface RouteProps {
-  parent: any;
+  parent?: boolean;
```
joseluisq: 👍

index.d.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/** location */
+interface LocationState {
+  pathname: string;
+  previous: string;
+}
+
+interface LocationActions {
+  go: (pathname: string) => void;
+  set: (data: LocationState) => LocationState;
```
Swizz: `set` is more like an internal action, I don't think we should expose it.

index.d.ts
```diff
@@ -0,0 +1,59 @@
...
+): VNode<RenderProps<P>> | undefined;
+
+/**Switch */
+export function Switch<P>(
+  props: object,
+  children: Array<VNode<RouteProps<P>>>
+): VNode<object>;
+
+/** Redirect */
+type RedirectProps = LinkProps;
```
Swizz: The `from` prop of `Redicrect` is missing in `LinkProps`

index.d.ts
```diff
@@ -0,0 +1,59 @@
...
+interface RouteProps<P> {
+  parent?: boolean;
+  path: string;
+  location?: Location;
+  render: (props: RenderProps<P>) => VNode<RenderProps<P>>;
+}
+
+export function Route<P>(
+  props: RouteProps<P>
+): VNode<RenderProps<P>> | undefined;
```
Swizz: `void` instead of `undefined` ?

index.d.ts
joseluisq: Since `from` __or__ `location` props could be optional
https://github.com/hyperapp/router/blob/8db6747b3598c5c7832ffe1c0f09311fecb0ceee/src/Redirect.js#L4

I suggest something like:

```ts
interface LinkProps {
  to: string
  location?: Location
}

interface RedirectProps extends LinkProps {
  from?: string
}
```

index.d.ts
m0a: thank you it fixed.

index.d.ts
m0a: thank you it fixed.

index.d.ts
m0a: thank you it fixed.

joseluisq:
It should be ready to merge 👍

kyegupov:
Any reason this is not merged yet? TypeScript is the only sane way to write frontend apps, this is a necessity.

jorgebucaran:
Hi, @kyegupov! I could merge it and publish a new patch, but I have no time to verify if it works and in case it breaks the router, then I'm afraid it will go on broken until I am done with https://github.com/jorgebucaran/hyperapp/pull/726. 😄

jorgebucaran:
🎉Just published 0.7.1 (0f5a002).
  
 cc @m0a @joseluisq @Swizz @kyegupov

joseluisq:
@jorgebucaran  :+1:

Issue #34, opened 2017-12-29
Before/After Router hooks


ansarizafar:
Please provide support for Before/After router hooks for router authorization.

JorgeBucaran:
And how would it look like in pseudocode? 🤔

ansarizafar:
Stenciljs router https://github.com/ionic-team/stencil-router has a routeRender function for route authorization.

JorgeBucaran:
What do you think @lukejacksonn @SkaterDad?

ansarizafar:
Any update on this issue?

JorgeBucaran:
Is anyone familiar enough with ReactRouter to know how they deal with this? This router's design is a lot like React's router, so it would make sense to check what they are doing and do it here as well, instead of coming up with something special for it.

JorgeBucaran:
@ansarizafar Can you provide pseudocode of how you think this would look like in Hyperapp? Please don't post another link.

Swizz:
Cant we use "fake" lifecycle events on `Route`/`Switch` component ?

```jsx
<Route path="/" render={Home}  before={beforeRoute} after={afterRoute}/>
```

PR #49, opened 2018-02-27
Create AllowRoutes.js

jonlov:
Hello everyone! 

I'm currently working in some authentification ith hyperapp and I needed to allow routes depending on an if statement and redirect if its false. I wrote this util for me, maybe should we add it to the router so anyone could easily import it.

The example usage will be:
```
<AllowRoutes if={!auth.user}>
    <Route path="/auth/login" render={Login(state, actions)}/>
    <Route path="/auth/signup" render={Signup(state, actions)}/>
</AllowRoutes>
```

and if you would like to redirect if the condition is not meet:
```
<AllowRoutes if={auth.user} redirectTo="/auth/login">
    <Route path="/auth/logout" render={Logout(state, actions)}/>
    <Route path="/user/settings" render={UserSettings(state, actions)}/>
</AllowRoutes>
```

Hope is helpful

### DIFF
--- a/src/AllowRoutes.js
+++ b/src/AllowRoutes.js
@@ -1,0 +1,17 @@
+import {h} from 'hyperapp';
+import {Redirect} from "./";
+
+export default({
+    if: _if,
+    redirectTo = '/'
+}, children) => {
+    const redirect = () => <Redirect to={redirectTo}/>;
+
+    if (!_if && window.location.pathname !== redirectTo && children.length)
+        redirect();
+
+    if (_if)
+        return children;
+
+    return null;
+}


jorgebucaran:
Hi, @jonlov. Do you know how ReactRouter handles this kind of thing?

jonlov:
@jorgebucaran they don't have any module to allow a route. You need to do it manually for each component or create a wrapper like this.

I'm not sure why they don't create a module for this. I always need to create it.

jorgebucaran:
@jonlov Where do we draw the line? I feel this module is very opinionated already. If you can solve this problem in userland, then let's solve it in userland.

jonlov:
@jorgebucaran It's optional, I shared here because I think is very helpful when doing routing permissions. It won't hurt anyone to have it ready to use but if not then we could just create another package just for this little feature.

jorgebucaran:
@jonlov Maybe hyperapp-router-extras if you plan to add a lot of stuff.

jonlov:
@jorgebucaran yes could be. That's the thing. I don't think there is much stuff for routes. However, I'm kind thinking to create a package for login/signup/forgot-password for passportJS with hyperapp and I guess I could add it there. It feels like is more for permissions on the routes.

On the other hand, ReactRouter has a [Prompt](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/Prompt.js) module. Which I think is very optional as well, but they add it anyways. I never used before.

jorgebucaran:
@jonlov What do you mean by optional? Every module is optional.

jonlov:
@jorgebucaran opinionated*

jorgebucaran:
@jonlov What if we add a section to the documentation in the mean time?

The router is going to get a makover someday soon after hyperapp/hyperapp#606.

jonlov:
@jorgebucaran sure let's add a section to the documentation then. I think this can also solve #34. What do you think?

PR #60, opened 2018-03-22
Don't pass through "to" and "location" attributes from <Link> to <a>

yuku-t:
image

### Note
There is a small breaking change:

```js
// No longer works
const sharedLinkProps = { to: "/path" }

h(Link, sharedLinkProps) // sharedLinkProps.to is deleted here
h(Link, sharedLinkProps)
h(Link, sharedLinkProps)
```


jorgebucaran:
I thought we had already done this once haha. 👍

PR #61, opened 2018-03-30
Custom Link.onclick handler

frenzzy:
The `onclick` attribute is not longer ignored and it allows you to prevent page transition depends on various conditions:
```js
<Link
  onclick={e => {
    if (condition) {
      // prevent page transition if condition is true
      e.preventDefault()
    }
  }}
/>
```

### DIFF
--- a/src/Link.js
+++ b/src/Link.js
@@ -4,12 +4,17 @@ export function Link(props, children) {
   return function(state, actions) {
     var to = props.to
     var location = state.location || window.location
+    var onclick = props.onclick
     delete props.to
     delete props.location
 
     props.href = to
     props.onclick = function(e) {
+      if (onclick) {
+        onclick(e)
+      }
       if (
+        e.defaultPrevented ||
         e.button !== 0 ||
         e.altKey ||
         e.metaKey ||
--- a/test/link.test.js
+++ b/test/link.test.js
@@ -103,3 +103,13 @@ test("pass through attributes", () => {
   expect(vnode.attributes.href).toEqual("/path")
   expect(vnode.attributes.pass).toEqual("through")
 })
+
+test("custom onclick handler", () => {
+  const event = { defaultPrevented: false }
+  const onclick = e => {
+    e.defaultPrevented = true
+  }
+  const vnode = h(Link, { to: "/path", onclick })({}, {})
+  vnode.attributes.onclick(event)
+  expect(event.defaultPrevented).toBe(true)
+})


jorgebucaran:
@frenzzy Why not use a regular `<a>` then?

About `preventDefault`, we could allow you to do it as a prop to the Link component instead:

```js
<Link preventDefault={true} />
```

Swizz:
Here you have both the `onclick` handler then the redirection.
And you can prevent the redirection behavior from the `onclick` handler

src/Link.js
```diff
@@ -4,12 +4,17 @@ export function Link(props, children) {
...
     delete props.to
     delete props.location
 
     props.href = to
     props.onclick = function(e) {
+      if (onclick) {
+        onclick(e)
+      }
       if (
+        e.defaultPrevented ||
```
jorgebucaran: @frenzzy I imagine this works because `e` is mutated when `e.preventDefault()` is run?

test/link.test.js
```diff
@@ -103,3 +103,15 @@ test("pass through attributes", () => {
   expect(vnode.attributes.href).toEqual("/path")
   expect(vnode.attributes.pass).toEqual("through")
 })
+
+test('custom onclick handler', () => {
```
jorgebucaran: Run `npm run format`.

test/link.test.js
```diff
@@ -103,3 +103,15 @@ test("pass through attributes", () => {
...
+  const event = {
+    preventDefault: jest.fn(),
```
jorgebucaran: @frenzzy Could you implement this without jest.fn()?

src/Link.js
jorgebucaran: Looks like it. 👍

test/link.test.js
frenzzy: Why?

test/link.test.js
Swizz: `npm run format -- 'test/**/*.js'`

The test directory is still not in the format config

jorgebucaran: The script is not covering the test files, whoops. Could you add it?

test/link.test.js
jorgebucaran: I want to avoid it for now.

test/link.test.js
frenzzy: PR: https://github.com/hyperapp/router/pull/62

jorgebucaran: 💯

frenzzy:
> Also, this needs to be documented.

The `Link` component is accept any event handlers and lifecycle attributes which hyperapp allows. Before this change it could be documented that you can't use `onclick` event handler and describe why. But now it behaves the same way as any other event handlers like `onmousedown` and so on. So, what you want to see in the documentation?

jorgebucaran:
@frenzzy Nevermind then, I'll write the docs, it's my turn to commit something too. 😉

jorgebucaran:
Thank you, @frenzzy! 🎉

PR #63, opened 2018-03-30
Fix Switch to handle lazy components

yuku-t:
Because `<Route>` is implemented as lazy component, `<Switch>`'s children argument is an array of functions.

### DIFF
--- a/test/switch.test.js
+++ b/test/switch.test.js
@@ -2,11 +2,15 @@ import { Switch } from "../src/Switch"
 
 test("Switch returns only the first truthy child", () => {
   const expected = "truthy value"
-  const children = [0, null, expected, false, "other"]
-  expect(Switch(null, children)).toBe(expected)
+  const children = [0, null, expected, false, "other"].map(child => () => child)
+  expect(Switch(null, children)()).toBe(expected)
 })
 
 test("Switch returns falsy when all children falsy", () => {
-  const children = [0, "", false, null]
-  expect(Switch(null, children)).toBeFalsy()
+  const children = [0, "", false, null].map(child => () => child)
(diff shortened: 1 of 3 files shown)


Swizz:
Oh nice catch, I was stucks on the fact that `Switch` doent need the state.

Can you fix the tests ? Or maybe make `Switch` aware of Lazy Component or not ?
So if `typeof children[i] == "function"`  this Lazy Component, else simple component

yuku-t:
I fixed the test. Then `yarn run format`

test/switch.test.js
```diff
@@ -2,11 +2,15 @@ import { Switch } from "../src/Switch"
 
 test("Switch returns only the first truthy child", () => {
   const expected = "truthy value"
-  const children = [0, null, expected, false, "other"]
-  expect(Switch(null, children)).toBe(expected)
+  const children = [0, null, expected, false, "other"].map(child => () => child)
+  expect(Switch(null, children)()).toBe(expected)
```
Swizz: Maybe we can make explicit the use of `state` and `actions` here ? 
```jsx
const state = {}
const actions = {}
expect(Switch(null, children)(state, actions)).toBe(expected)
```
Just thinking about  readability and understanding of the use of Lazy Component.

yuku-t:
Fixed test files. It also improves code coverage 📈

jorgebucaran:
Thank you, @yuku-t! 🎉

yuku-t:
@jorgebucaran I hope that this PR will be released as soon as possible 🙏

jorgebucaran:
@yuku-t Sorry for the delay. It's available now as 0.5.1.

PR #64, opened 2018-04-02
Remove broken `state first` feature

yuku-t:
Since hyperapp/router relies on History API especially "popstate" event, developers must add an event listener to `window`. In general, calling `location.subscribe()` acts for the purpose, but the function needs `location.actions` as an argument. This means that developers must wire `location` slice to `actions`, and to make matters worse `location.actions` updates `state.location` not `window.location` so hyperapp/router does not use the `window.location` as source of truth in this situation.

Consequently, there is no way to skip wiring the `location` slice to `state` and `actions`.

Another solusion is to provide `location.subscribeWindowWithoutActions()` but, in my opinion, it should be implemented as another npm package.


jorgebucaran:
I think we introduced a bug [here](https://github.com/hyperapp/router/pull/59/files#diff-5dcf83f7c58c2984faf518d25944c39bR6) by using `state.location` instead of `props.location`.

> Consequently, there is no way to skip wiring the location slice to state and actions.

Yes, to use the router we need to wire the router's `state` and `actions` slice as usual — what I want to avoid is passing `window.location` as a prop to every component.

jorgebucaran:
Available as 0.6.0. 👌

jorgebucaran:
Thank, you @yuku-t! 🎉 🎉

Swizz:
I added this functionality only for "compatibility" purpose, I mean to let the use of the previous behavior.

PR #65, opened 2018-04-02
Improve tests

yuku-t:
- Use JSX for better code readability.
- Dispatch a native `MouseEvent` instead of calling `vnode.attributes.onclick`.
- Move all integration tests (which contain `app()` in their test code) to test/index.test.js.
    - In previous, test/link.test.js imported `Route`, `Link` and `location`.
    - Now, it imports `Link` only. It concentrates on unit testing.
- 💯 test coverage.
- Fix potential bug around external links.


test/route.test.js
```diff
@@ -46,3 +46,16 @@ test("Route without path prop matches to every location", () => {
...
   expect(Route({ render }, [])({ location: { pathname: "/users" } })).toBe(true)
 })
+
+test("Route ignores url params containing invalid character sequences", () => {
+  const invalidChars = "%E0%A4%A"
+  expect(() => decodeURI(invalidChars)).toThrow(URIError)
+  const state = { location: { pathname: `/foo/${invalidChars}/bar/baz` } }
+  const actions = {}
+  const render = jest.fn(({ match }) => {
+    expect(match.params).toEqual({ bar: "baz" })
```
yuku-t: @jorgebucaran Is this intended behavior? In my opinion, a malformed URI should not match to any routes. What do you think?

frenzzy: Why not `toEqual({ foo: "%E0%A4%A", bar: "baz" })`

yuku-t: Removing `decodeURI` from ↓ changes `match.params` to `{ foo: "%E0%A4%A", bar: "baz" }`. I'm not sure why `decodeURI` is used here.

https://github.com/hyperapp/router/blob/795eb9b81ad9a6027d74dc65b0fe9548b0470095/src/parseRoute.js#L30-L34

frenzzy: Decoding is make sense because you usually expect `{ name: "café" }` instead of `{ name: "caf%C3%A9" }` in the params object.

but probably the code above should looks like this:
```js
function decodeParam(val) {
  try {
    return decodeURIComponent(val)
  } catch (err) {
    return val
  }
}

params[paths[i].slice(1)] = urls[i] = decodeParam(urls[i])
```

package.json
```diff
@@ -27,12 +27,24 @@
...
   "babel": {
-    "presets": "env"
+    "presets": [
+      "env"
+    ],
+    "plugins": [
+      [
+        "transform-react-jsx",
+        {
+          "pragma": "h"
```
jorgebucaran: 👍

test/route.test.js
jorgebucaran: > @yuku-t In my opinion, a malformed URI should not match to any routes. What do you think?

A lot of code was golfed to make the router as small as possible. It was fun, but probably a waste of time. I am too obsessed about trading bytes for preciseness anymore and I think it's time to revisit a lot of the implementation code.

package.json
```diff
@@ -27,12 +27,24 @@
...
+        }
+      ]
+    ]
   },
   "jest": {
     "testURL": "http://localhost"
   },
   "devDependencies": {
+    "babel-jest": "^22.4.3",
+    "babel-plugin-transform-react-jsx": "^6.24.1",
```
jorgebucaran: @yuku-t I don't expect you to write all tests to use JSX, but could you use JSX in these tests? 🙏

test/link.test.js
```diff
@@ -1,74 +1,18 @@
...
 
-test("custom onclick handler", () => {
-  const event = { defaultPrevented: false }
-  const onclick = e => {
-    e.defaultPrevented = true
-  }
-  const vnode = h(Link, { to: "/path", onclick })({}, {})
+test("Calling onclick of VNode transparently calls Link's onclick prop", () => {
+  const onclickProp = jest.fn()
+  const vnode = Link({ onclick: onclickProp })({}, {})
```
jorgebucaran: JSX

test/route.test.js
```diff
@@ -46,3 +46,16 @@ test("Route without path prop matches to every location", () => {
...
+  })
+  expect(() =>
+    Route({ path: "/foo/:foo/bar/:bar", render }, [])(state, actions)
```
jorgebucaran: JSX here too. 🙏

test/link.test.js
yuku-t: It is intentional design. There are two reasons:

1. Because of lazy component, the mentioned line equals to the following. Don't you feel strange?

    ```js
    const vnode = (<Link onclick={onclickProp} />)({}, {})
    ```

2. test/link.test.js is for unit testing. `Link` could be used as a component, but it is a pure function.
    - On the other hand, test/index.test.js is for integration testing, so I used JSX there.

yuku-t:
- Fix `<Link>` as https://github.com/hyperapp/router/pull/65#discussion_r178557646
- Add test scenario
- Fix potential bug around external link

test/link.test.js
jorgebucaran: Okay, makes sense. That would be weird indeed.

test/index.test.js
```diff
@@ -1,38 +1,124 @@
...
+import { Route, Link, Redirect, location } from "../src"
 
-test("Router", done => {
-  const state = {
-    location: location.state
-  }
+const wait = async ms => new Promise(resolve => setTimeout(resolve, ms))
+const click = e => e.dispatchEvent(new MouseEvent("click", { button: 0 }))
+
+let state, actions, unsubscribe
```
jorgebucaran: I try to avoid using `var` and `let` when writing pure ES6, but I can fix this later!

PR #68, opened 2018-04-18
Use CustomEvent polyfill if necessary

### LINKED ISSUE #58: CustomEvent not support IE
when I use in IE(11 and below), it fails with:
```
location.js (7,1)  Object doesn't support this action
```

yuku-t:
Close #58

To support IE11

## Reference

https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "history"
   ],
   "scripts": {
-    "test": "jest --coverage --no-cache",
+    "test": "jest --coverage --no-cache --collectCoverageFrom=src/**/*.js --collectCoverageFrom=!src/CustomEvent.js",
     "build": "npm run bundle && npm run minify",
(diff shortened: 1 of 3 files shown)


yuku-t:
Because there is `window.CustomEvent` in jest env, I cannot keep 100% coverage 😭

jorgebucaran:
@yuku-t Why do we need to include the polyfill in the router? Surely, it must be possible to polyfill `CustomEvent` in the environment before importing @hyperapp/router.

yuku-t:
You are right, the dependency should be noted in the README.md though

jorgebucaran:
Yeah, that's more reasonable. I'll add it later. Thanks!

PR #70, opened 2018-04-25
BasePath

kidandcat:
Include Basepath

### DIFF
--- a/src/Route.js
+++ b/src/Route.js
@@ -2,7 +2,19 @@ import { parseRoute } from "./parseRoute"
 
 export function Route(props) {
   return function(state, actions) {
-    var location = state.location
+    var location = state.location;
+    // Basepath must end with /
+    if(props.basepath && props.basepath[props.basepath.length] !== "/"){
+      props.basepath += "/";
+    }
+    // Remove basepath from path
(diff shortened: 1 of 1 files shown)


Swizz:
What is the benefit of the `basepath` property comparing to this simple concatenation ?
```jsx
const basepath = "demo"
// ...
<Route path={`${basepath}/topic`} render={Topic} />
```

kidandcat:
I'm having a separate hyperapp app as a widget: https://github.com/kidandcat/newsuncork/blob/master/src/apps/product-creator.jsx

And I'm instancing it here: https://github.com/kidandcat/newsuncork/blob/master/src/components/CreateProduct.jsx

I'll investigate why today, but when it loads, the location of CreateProduct.jsx's router is "/" (but the URL path is /admin/product/create)

jorgebucaran:
@kidandcat We should add features to the Router which are impossible to add from the outside. Couldn't you solve this problem in userland?

kidandcat:
I couldn't do it in the userland, but I managed to do it in a much more cleaner way:

https://github.com/hyperapp/router/pull/72

@jorgebucaran

PR #72, opened 2018-05-02
Allow multiple routers interaction

kidandcat:
Allow multiple routers to coexist by not wrapping the pushstate and replacestate functions multiple times and triggering the router listeners after creating it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
-  "name": "@hyperapp/router",
+  "name": "hyperapp-patched-router",
   "description": "Declarative routing for Hyperapp using the History API.",
   "version": "0.7.0",
   "main": "dist/router.js",
   "module": "src/index.js",
   "license": "MIT",
-  "repository": "hyperapp/router",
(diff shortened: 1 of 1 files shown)


src/location.js
```diff
@@ -2,6 +2,11 @@ function wrapHistory(keys) {
   return keys.reduce(function(next, key) {
     var fn = history[key]
 
+    // Do not wrap if it's already wrapped
+    if(history[key].toString().indexOf('CustomEvent("pushstate"') !== -1){
+      return function(){};
+    }
```
frenzzy: I guess wrapping up a native history object was a bad idea initially...

kidandcat: Well, it works well like it is now, but the thing is, you can't use the router 2 times completely separated, because you will wrap the object two times, then you will have 2 copies of every event.

kidandcat:
I still need this change, and I need to rewrite it in node_modules everytime I deploy, so I'm going to fork and publish it to npm

PR #73, opened 2018-05-12
IE 11 also makes expected behavior

atomita:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
Does not work with IE11.

## What is the new behavior?
IE 11 also makes expected behavior.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

### DIFF
--- a/src/Link.js
+++ b/src/Link.js
@@ -1,7 +1,10 @@
 import { h } from "hyperapp"
 
 function getOrigin(loc) {
-  return loc.protocol + "//" + loc.hostname + (loc.port ? ":" + loc.port : "")
+  var protocol = (loc.protocol && ':' !== loc.protocol) ? loc.protocol : location.protocol
+  var host = [loc.hostname, loc.port].filter(Boolean).join(":")
+      || [location.hostname, location.port].filter(Boolean).join(":")
+  return protocol + "//" + host
 }
 
 function isExternal(anchorElement) {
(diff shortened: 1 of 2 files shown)


src/Link.js
```diff
@@ -1,7 +1,10 @@
 import { h } from "hyperapp"
 
 function getOrigin(loc) {
-  return loc.protocol + "//" + loc.hostname + (loc.port ? ":" + loc.port : "")
+  var protocol = (loc.protocol && ':' !== loc.protocol) ? loc.protocol : location.protocol
```
jorgebucaran: @atomita Could you please explain what's going on here? We should create a function to document whatever bug of browser inconsistency this is trying to patch.

atomita: @jorgebucaran 
I reconfirmed this issue.
May be a issue of access by localhost or IP.
I confirmed that `document.getElementsByTagName('a')[0].hostname` will be empty if I open host (http: 10.0.2.2: 9081 or http://localhost:9081) on IE 11 from windows 7 using "Modern.IE".

But, trying a small code with planker, the protocol and hostname seem to be fine. (Port has a issue)
http://plnkr.co/edit/PRtqM2AW0Daq3CbJvSO8

PR #92, opened 2018-12-02
Only way found to fix nasty erros on TS build.

nmpribeiro:
- Reason for `export function Link(props: LinkProps): VNode<LinkProps> | any;` change request:
```
    TS2605: JSX element type 'void | VNode<LinkProps>' is not a constructor function for JSX elements.
  Type 'void' is not assignable to type 'Element | null'.
```

- Reason for `export function Route<P>(  props: RouteProps<P>): VNode<RenderProps<P>> | any;` change request:
```
    TS2605: JSX element type 'VNode<RenderProps<{}>>' is not a constructor function for JSX elements.
  Type 'VNode<RenderProps<{}>>' is missing the following properties from type 'Element': type, props
```

Let me know if you guys find exactly which type we should declare.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ interface LinkProps {
   to: string;
   location?: Location;
 }
-export function Link(props: LinkProps): VNode<LinkProps>;
+export function Link(props: LinkProps): VNode<LinkProps> | any;
 
 /** Route */
 interface Match<P> {
@@ -28,7 +28,7 @@ interface RouteProps<P> {
 
 export function Route<P>(
   props: RouteProps<P>
-): VNode<RenderProps<P>> | void;
+): VNode<RenderProps<P>> | any;
 
 /**Switch */
 export function Switch<P>(


Swizz:
The first warning seem explicitly enough for me. I would move on VNode<LinkProps> | null.
A VNode must not be undefined but can be null.
For the second, I guess `Hyperapp.VNode` should extends `JSX.Element`.

Plus, returns seem incorrect, as since #59 helpers do not return VNode, they return Lazy Components.

nmpribeiro:
@Swizz 
```
declare global {
  namespace JSX {
    interface Element extends VNode<any> {}
    interface IntrinsicElements {
      [elemName: string]: any
    }
  }
}
```
Tested null for the first and it didn't work as well...
@Swizz have you tested your suggestions?

Swizz:
Untested. I knew that DT use `JSX.Element | null` on some places : 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8ff7e8f29e059f3799c19e0bafb7ac010786ca50/types/react/v15/index.d.ts#L314

I think it is more related to `Element` and not to the `null` as `any` match with `Element | null`.

So I would try  to extend `Hyperapp.VNode` from `JSX.Element`.

nmpribeiro:
But this means an update on the `hyperapp` dependency. What's your take on this @jorgebucaran ?

jorgebucaran:
@nmpribeiro Maybe someone with more TS experience can comment.

nmpribeiro:
At least someone able to understand the return types of these JSX.Element and the VNode. h function should only expect JSX.Element right (at least only VNode)?. I will look at that again when possible.

PR #93, opened 2018-12-06
[TS] Add class to LinkProps type

niicojs:
Having a class is a common scenario which gives me error all the time.
Simple fix for that.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,6 +4,7 @@ import { VNode } from "hyperapp";
 interface LinkProps {
   to: string;
   location?: Location;
+  class?: string;
 }
 export function Link(props: LinkProps): VNode<LinkProps>;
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -4,6 +4,7 @@ import { VNode } from "hyperapp";
 interface LinkProps {
   to: string;
   location?: Location;
+  class?: string;
```
frenzzy: maybe `[attributeName]: any` ?

niicojs: I don't need it so didn't implement it. I don't know a way to be exhaustive here.

slacktracer:
Sooo... Can we have it? =)

jorgebucaran:
I'm having a technical issue and couldn't publish the package, but I've already contacted npm and asked them for help. A new bump should be ready to ship in a few hours, give or take. 👋

PR #97, opened 2018-12-15
Fix TS2605 Error on `Route`

### LINKED ISSUE #96: TS2605 Error appeared on TypeScript 3.2.2
On TypeScript 3.2.2, every &lt;Route&gt; outputs TS2605 error.

like this:
```
JSX element type 'void | VNode<RenderProps<{}>>' is not a constructor function for JSX elements.
  Type 'void' is not assignable to type 'Element | null'.ts(2605)
```

babie:
Fix #96

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,7 +28,7 @@ interface RouteProps<P> {
 
 export function Route<P>(
   props: RouteProps<P>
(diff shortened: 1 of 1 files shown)


jorgebucaran:
Thanks, @babie! 👋

PR #106, opened 2019-05-30
Proper typing for LinkProps.

slacktracer:
TS compiler complains without it... So I wasn't being able to add a class to my links as I learned in https://github.com/jorgebucaran/hyperapp-router/issues/86.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,6 +2,7 @@ import { Children, VNode } from "hyperapp";
 
 /** Link */
 interface LinkProps {
+  class?: string;
   to: string;
   location?: Location;
 }


frenzzy:
maybe something like this is better?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4685bbd27150e7261dd65705b72cc812f9651b43/types/react-router-dom/index.d.ts#L49-L53

frenzzy:
This is the same as #93

slacktracer:
@frenzzy Actually it's even better as it accounts for all props. Just found out looking the code that all props are passed down...

